PR #12, opened 2018-01-16
Tika

mgax:


snoop/data/analyzers/tika.py
```diff
@@ -0,0 +1,71 @@
+from urllib.parse import urljoin
+from django.conf import settings
+import requests
+from ..tasks import shaorma
+from .. import models
+
+TIKA_CONTENT_TYPES = [
+    'application/pdf',
```
gabriel-v: also add `text/plain` because emails

PR #13, opened 2018-01-17
detect eml files by .split(':')[0].strip().title()

gabriel-v:


snoop/data/magic.py
```diff
@@ -82,3 +82,23 @@ def download_magic_definitions():
...
+def looks_like_email(path):
+    HEADER_SET = {
+        "Relay-Version" , "Return-Path" , "From" , "To" ,
+        "Received" , "Message-Id" , "Date", "In-Reply-To", "Subject"
+    }
+    HEADER_MIN_HIT_COUNT = 2
+    HEADER_READ_SIZE = 1024 * 64
+
+    with path.open('rb') as f:
+        content = f.read(HEADER_READ_SIZE).decode('latin-1')
```
mgax: Yeaaahhh... there's this edge case where you [might](https://docs.python.org/3/library/io.html#io.RawIOBase.read) read less than `HEADER_READ_SIZE`, even if there is more data.

gabriel-v:
it's `message/rfc822` not `message/rfc-822`

snoop/data/models.py
```diff
@@ -83,6 +83,17 @@ def create(cls):
         pk = fields.pop('sha3_256')
 
         Path(f.name).rename(BLOB_ROOT / pk)
+
+        if fields['mime_type'].startswith('text/'):
+            if looks_like_email(Path(BLOB_ROOT / pk)):
```
mgax: Let's not repeat `Path(BLOB_ROOT / pk)`, that's liable to change pretty soon, if we want to store millions of blobs. At least save it into a variable.

PR #19, opened 2018-01-17
also unarchive pst files

gabriel-v:


snoop/data/analyzers/archives.py
```diff
@@ -7,16 +7,43 @@
...
+    'application/x-tar',
+}
+
+READPST_KNOWN_TYPES = {
+    'application/x-hoover-pst',
 }
 
 
+def is_archive(mime_type):
+    return mime_type in SEVENZIP_KNOWN_TYPES + READPST_KNOWN_TYPES
```
mgax: For the sake of optimization, let's concatenate them once, and save the result to a global.

PR #21, opened 2018-01-17
Make the digest admin great again

mgax:


snoop/data/admin.py
```diff
@@ -1,9 +1,14 @@
 from django.urls import reverse
 from django.contrib import admin
 from django.utils.safestring import mark_safe
+from django.template.defaultfilters import truncatechars
 from . import models
 
 
+def short(blob_pk):
+    return truncatechars(blob_pk, 32)
```
gabriel-v: make it even shorter, i'm getting tired of looking at walls of hashes... or maybe show `aabbcc...dd` because that's how you compare it manually anyway

PR #23, opened 2018-01-18
Save first shaorma blob argument as foreign key

mgax:


snoop/data/analyzers/text.py
```diff
@@ -2,7 +2,7 @@
 from ..tasks import shaorma
 
 
-@shaorma
+@shaorma('text.extract')
 def extract_text(blob_pk):
```
gabriel-v: why not turn this arg into `blob` too?

snoop/data/filesystem.py
```diff
@@ -68,37 +68,38 @@ def file_to_blob(directory, name):
...
     else:
         digest_blob = file.blob
 
-    digest.laterz(file.collection.pk, digest_blob.pk, depends_on=depends_on)
+    digest.laterz(digest_blob, file.collection.pk, depends_on=depends_on)
 
 
-@shaorma
+@shaorma('filesystem.create_archive_files')
 def create_archive_files(file_pk, archive_listing):
```
gabriel-v: `archive_listing` is a blob, reverse the arguments to save it to the task

snoop/data/filesystem.py
```diff
@@ -153,8 +154,8 @@ def create_file(parent_directory, name, blob):
     create_directory_children(fake_root, archive_listing_data)
 
 
-@shaorma
-def digest(collection_pk, blob_pk, **depends_on):
+@shaorma('filesystem.digest')
+def digest(blob_pk, collection_pk, **depends_on):
```
gabriel-v: this should be `blob`, not the `blob_pk`, right? maybe the tests pass because Django's ORM does smart things with your `pk=Blob` filter

testsuite/test_shaorma.py
```diff
@@ -22,3 +22,23 @@ def two(one_result):
...
+
+        return output.blob
+
+    with models.Blob.create() as writer:
+        writer.write(b'hello')
+
+    task = with_blob.laterz(writer.blob, 'world')
+    task.refresh_from_db()
+    with task.result.open() as f:
+        assert f.read() == b'hello world'
```
gabriel-v: also test that `task.blob_arg == writer.blob` in this or some other test

snoop/data/filesystem.py
mgax: Turns out `archive_listing` comes from `depends_on` so it's already a blob foreign key from the `TaskDependency` model.

PR #26, opened 2018-01-19
Cleanup

mgax:


snoop/data/tasks.py
```diff
@@ -114,3 +118,16 @@ def laterz(*args, depends_on={}):
...
 
     return decorator
+
+
+def dispatch_pending_tasks():
+    for task in models.Task.objects.filter(status=models.Task.STATUS_PENDING):
+        deps_not_ready = (
+            task.prev_set
+            .exclude(prev__status=models.Task.STATUS_SUCCESS)
+            .count()
```
gabriel-v: `.exists()`: https://docs.djangoproject.com/en/2.0/ref/models/querysets/#django.db.models.query.QuerySet.exists

PR #27, opened 2018-01-19
Originals

mgax:
Keep original blobs in `File.original` and use `File.blob` for mangled versions of the data, e.g. when converting `.emlx` or `.msg` files to `.eml`.


snoop/data/filesystem.py
```diff
@@ -73,34 +73,36 @@ def handle_file(file_pk):
...
+        unarchive_task = archives.unarchive.laterz(file.original)
         create_archive_files.laterz(
             file.pk,
             depends_on={'archive_listing': unarchive_task},
         )
 
-    elif file.blob.mime_type == "application/vnd.ms-outlook":
-        digest_blob = email.msg_blob_to_eml(file.blob)
+    elif file.original.mime_type == "application/vnd.ms-outlook":
+        digest_blob = email.msg_blob_to_eml(file.original)
```
gabriel-v: `file.blob`?

snoop/data/filesystem.py
```diff
@@ -112,8 +114,8 @@ def create_directory_children(directory, children):
         child_files = []
         for item in children:
             if item['type'] == 'file':
-                blob = models.Blob.objects.get(pk=item['blob_pk'])
-                file = create_file(directory, item['name'], blob)
+                original = models.Blob.objects.get(pk=item['blob_pk'])
```
gabriel-v: this function should still work on `File.blob`, in case we want to mangle archives somehow

snoop/data/filesystem.py
```diff
@@ -73,34 +73,36 @@ def handle_file(file_pk):
     file = models.File.objects.get(pk=file_pk)
     depends_on = {}
 
-    if archives.is_archive(file.blob.mime_type):
-        digest_blob = file.blob
-        unarchive_task = archives.unarchive.laterz(file.blob)
+    if archives.is_archive(file.original.mime_type):
+        file.blob = file.original
+        unarchive_task = archives.unarchive.laterz(file.original)
```
gabriel-v: the `unarchive` operation should happen on the mangled `File.blob`, not the original

testsuite/test_archives.py
```diff
@@ -140,7 +140,7 @@ def test_create_archive_files():
         ctime=filesystem.time_from_unix(0),
         mtime=filesystem.time_from_unix(0),
         size=0,
-        blob=zip_blob,
+        original=zip_blob,
```
gabriel-v: also set `blob=zip_blob` because the unarchive functions should look at that field

PR #28, opened 2018-01-22
Api features

mgax:
* show full path to a file
* click-browse through directories
* paginate the api document feed
* download original files

### DIFF
--- a/snoop/data/digests.py
+++ b/snoop/data/digests.py
@@ -74,6 +74,22 @@ def full_path(file):
     return '/'.join(reversed(elements))
 
 
+def directory_id(directory):
+    return f'_directory_{directory.pk}'
+
+
+def parent_id(file):
+    parent = file.parent
+
+    if isinstance(parent, models.File):
+        return parent.blob.pk
+
+    if isinstance(parent, models.Directory):
+        return directory_id(parent)
+
+    return None
+
+
 def get_document_data(digest):
     with digest.result.open() as f:
         digest_data = json.loads(f.read().decode('utf8'))
@@ -82,6 +98,7 @@ def get_document_data(digest):
 
     return {
         'id': digest.blob.pk,
+        'parent_id': parent_id(first_file),
         'version': zulu(digest.date_modified),
         'content': {
             'content-type': digest.blob.mime_type,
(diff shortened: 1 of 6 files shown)


snoop/data/models.py
```diff
@@ -66,6 +66,13 @@ class Blob(models.Model):
     def __str__(self):
         return self.pk
 
+    @property
+    def content_type(self):
+        if self.mime_type.startswith('text/'):
+            return f"{self.mime_type}; charset={self.mime_encoding}"
+
+        return self.mime_type
```
gabriel-v: why not always include the `charset`? or include it when it isn't `binary`?

snoop/data/digests.py
```diff
@@ -65,15 +65,40 @@ def filetype(mime_type):
...
 def get_document_data(digest):
     with digest.result.open() as f:
         digest_data = json.loads(f.read().decode('utf8'))
 
     first_file = digest.blob.file_set.order_by('pk').first()
-    filename = path = first_file.name
 
     return {
         'id': digest.blob.pk,
+        'parent_id': parent_id(first_file),
```
gabriel-v: if the "Parent" button only leads you to the first file, then how can you navigate to the other places where a document shows up?

snoop/data/urls.py
```diff
@@ -1,8 +1,11 @@
 from django.urls import path
 from . import views
 
+
 urlpatterns = [
     path('<name>/json', views.collection),
     path('<name>/feed', views.feed),
+    path('<name>/_directory:<int:pk>/json', views.directory),
```
gabriel-v: please don't put `:` in URLs because some browsers might behave badly

snoop/data/digests.py
mgax: You can't yet.

snoop/data/models.py
mgax: I think it makes no sense unless the content is text. Do you have an example of a non-text mime type that has a charset? Or some documentation somewhere?

mgax:
> This breaks the "Parent" button.

Au contraire, it makes the parent button work. I've been testing with the UI.

PR #29, opened 2018-01-23
Emails

mgax:
Headers, attachments, șaorma cu de toate.

### DIFF
--- a/snoop/data/analyzers/email.py
+++ b/snoop/data/analyzers/email.py
@@ -8,6 +8,8 @@ from .. import models
 from ..tasks import shaorma, ShaormaError
 from . import tika
 
+BYTE_ORDER_MARK = b'\xef\xbb\xbf'
+
 
 def iter_parts(message, numbers=[]):
     if message.is_multipart():
@@ -72,7 +74,7 @@ def parse(blob):
     with blob.open() as f:
         message_bytes = f.read()
 
-    if message_bytes[:3] == b'\xef\xbb\xbf':
+    if message_bytes[:3] == BYTE_ORDER_MARK:
         message_bytes = message_bytes[3:]
 
     message = email.message_from_bytes(message_bytes)
--- a/snoop/data/analyzers/email.py
+++ b/snoop/data/analyzers/email.py
@@ -48,6 +48,22 @@ def dump_part(message):
     if message.is_multipart():
         rv['parts'] = [dump_part(part) for part in message.get_payload()]
 
+    if message.get_content_disposition():
+        raw_filename = message.get_filename()
+        if raw_filename:
+            filename = str(email.header.make_header(
+                email.header.decode_header(raw_filename)
+            ))
+            attachment_content = message.get_payload(decode=True)
+
+            with models.Blob.create() as writer:
+                writer.write(attachment_content)
+
+            rv['attachment'] = {
+                'name': filename,
+                'blob_pk': writer.blob.pk,
+            }
+
     return rv
 
 
--- a/snoop/data/analyzers/email.py
+++ b/snoop/data/analyzers/email.py
@@ -6,6 +6,7 @@ from collections import defaultdict
 import email
 from .. import models
 from ..tasks import shaorma, ShaormaError
+from . import tika
 
 
 def iter_parts(message, numbers=[]):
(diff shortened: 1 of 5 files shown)


snoop/data/analyzers/email.py
```diff
@@ -26,14 +27,56 @@ def get_headers(message):
...
+        payload_bytes = message.get_payload(decode=True)
+        with models.Blob.create() as writer:
+            writer.write(payload_bytes)
+        rmeta_blob = tika.rmeta(writer.blob)
+        with rmeta_blob.open(encoding='utf8') as f:
+            rmeta_data = json.load(f)
+        rv['text'] = rmeta_data[0]['X-TIKA:content']
+
+    if message.is_multipart():
+        rv['parts'] = [dump_part(part) for part in message.get_payload()]
```
gabriel-v: If you recurse into `dump_part` for parts of the message, `dump_part` will try to extract the headers from those: see the first line from `dump_part`. I don't think it's a good idea to look for headers where there shouldn't be any.

snoop/data/analyzers/email.py
```diff
@@ -26,14 +27,56 @@ def get_headers(message):
...
+    return rv
+
+
 @shaorma('email.parse')
 def parse(blob):
     with blob.open() as f:
-        message = email.message_from_bytes(f.read())
+        message_bytes = f.read()
+
+    if message_bytes[:3] == b'\xef\xbb\xbf':
```
gabriel-v: What is this? Please include the words `byte order mark, thanks Microsoft` in a comment above this.

snoop/data/filesystem.py
```diff
@@ -139,3 +146,47 @@ def create_file(parent_directory, name, original):
...
+        for attachment in attachments:
+            original = models.Blob.objects.get(pk=attachment['blob_pk'])
+            size = original.path().stat().st_size
+            now = timezone.now()
+
+            (file, _) = attachments_dir.child_file_set.get_or_create(
+                name=attachment['name'],
+                defaults=dict(
+                    collection=file.collection,
+                    ctime=now,
```
gabriel-v: I think these (`ctime`, `mtime`) should be a time taken from the email, not the current digest time. Or, if we can figure out their actual create/modify times from the email itself? If we can't figure them out, then just inherit them from the email file itself (the first one? the last one?)

snoop/data/digests.py
```diff
@@ -90,46 +90,88 @@ def parent_id(file):
...
+        if attachments:
+            rv['children'] = [
+                child_file(f)
+                for f in attachments.child_file_set.order_by('pk').all()
+            ]
+
+    return rv
+
+
+def child_file(file):
```
gabriel-v: There's nothing specific to a `child` file here: maybe name it `file_to_dict` or something?

snoop/data/digests.py
```diff
@@ -90,46 +90,88 @@ def parent_id(file):
...
-        }
-
-    def child_dir(directory):
-        return {
-            'id': directory_id(directory),
-            'content_type': 'application/x-directory',
-            'filename': directory.name,
-        }
 
+def child_dir(directory):
```
gabriel-v: There's nothing specific to a `child` dir here: maybe name it `dir_to_dict` or something?

snoop/data/digests.py
```diff
@@ -90,46 +90,88 @@ def parent_id(file):
...
+
+    text_bits = []
+    for part in iter_parts(email_data):
+        part_text = part.get('text')
+        if part_text:
+            text_bits.append(part_text)
+
+    headers_to = []
+    for header in ['To', 'Cc', 'Bcc', 'Resent-To', 'Recent-Cc', 'Reply-To']:
+        headers_to += headers.get(header, [])
```
gabriel-v: maybe deduplicate these? And `Reply-to` isn't really a "to" field.

testsuite/test_integration.py
```diff
@@ -58,5 +58,4 @@ def feed_page(url):
 
     # .partial.emlx
     partialemlx = docs[ID['partialemlx']]
-    headers = partialemlx['content']['_emailheaders']
-    assert headers['Subject'] == ["Re: promulgare lege"]
+    assert partialemlx['content']['subject'] == "Re: promulgare lege"
```
gabriel-v: could you also add a few of the above tested mails (maybe the most complex one) here in the integration test?

snoop/data/analyzers/email.py
mgax: We're going to need headers of sub-parts when we're going to decrypt pgp emails.

snoop/data/digests.py
mgax: It's specific to the "child" data structure which is part of the digest document:

```json
{
  "id": "foo",
  "content": {},
  "children": [{"id": "bar", "content_type": "text/plain", "filename": "bar.txt"}]
}
```

PR #30, opened 2018-01-25
Add missing dependency for a running task

mgax:
A task gets the ability to say "hey, I need this task to run before I can finish my work". It aborts execution, gets marked with state "pending", and the new task gets added to its `depends_on` list.

I've used this mechanism to shaormify email html-to-text processing, conversion from msg to eml, and reconstructing an emlx from parts. Maybe this is a good way to allow emlx reconstruction to wait for dependent files to be imported as blobs.

### DIFF
--- a/snoop/data/analyzers/email.py
+++ b/snoop/data/analyzers/email.py
@@ -94,6 +94,7 @@ def parse(blob, **depends_on):
     return output.blob
 
 
+@shaorma('email.msg_to_eml')
 def msg_to_eml(blob):
     with tempfile.TemporaryDirectory() as temp_dir:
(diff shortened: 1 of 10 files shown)


testsuite/settings.py
```diff
@@ -7,6 +7,3 @@
 SNOOP_TESTDATA = os.getenv('SNOOP_TESTDATA', default_testdata_path)
 
 assert os.path.isdir(SNOOP_TESTDATA)
-
-CELERY_TASK_ALWAYS_EAGER = True
```
gabriel-v: leave these here, we don't want to accidentally hit celery

PR #32, opened 2018-01-26
Fix race conditions related to task concurrency

mgax:


snoop/data/tasks.py
```diff
@@ -25,6 +25,19 @@ def __init__(self, name, task):
...
+    def send_to_celery():
+        laterz_shaorma.apply_async((task.pk,), queue=task.func)
+
+    transaction.on_commit(send_to_celery)
+
+
+def queue_next_tasks(task):
+    for next_dependency in task.next_set.all():
+        next = next_dependency.next
+        queue_task(next)
```
gabriel-v: `queue_task(next_dependency.next)` because `next` is a keyword

PR #34, opened 2018-01-29
One shaorma per file

mgax:
Import each shaorma into its own file, so that we isolate errors better. In particular, errors that come from bugs in the magic `file` program.

This required a change in `emlx.reconstruct` - it needs to wait for `walk` and `walk_file` to finish on `.emlxpart` files in its folder. This approach is going to be useful if we ever want to detect emlx attachments in the `Attachments` folder, which is one level up the tree, next to the `Messages` folder.

So I took the opportunity to extract a `require_dependency` helper that reads from `depends_on` and raises `MissingDependency` if needed.


snoop/data/filesystem.py
```diff
@@ -76,18 +75,16 @@ def handle_file(file_pk, **depends_on):
...
 
     elif file.original.mime_type == 'message/x-emlx':
-        eml = depends_on.get('emlx_reconstruct')
-        if not eml:
-            task = emlx.reconstruct.laterz(file.pk)
-            raise MissingDependency('emlx_reconstruct', task)
-        file.blob = eml
+        file.blob = require_dependency(
+            'emlx_reconstruct', depends_on,
+            lambda: emlx.reconstruct.laterz(file.pk),
```
gabriel-v: ugly but I understand

snoop/data/analyzers/emlx.py
```diff
@@ -18,13 +20,21 @@ def reconstruct(file_pk):
...
             part_name = re.sub(r'\.partial\.emlx$', ext, file.name)
+            parent = file.parent_directory
+
+            require_dependency(
+                f'walk-{parent.pk}', depends_on,
+                lambda: filesystem.walk.laterz(parent.pk),
+            )
+
+            require_dependency(
+                f'walk_file-{parent.pk}-part-{ext}', depends_on,
```
gabriel-v: Instead of `-part-{ext}` maybe put in the actual filename? Or say `-emlxpart` instead of `-part` so it's obvious it's about emlx files?

mgax: `emlxpart-` works. I'd rather not use the filename, who knows what scary stuff is in there.

snoop/data/analyzers/emlx.py
```diff
@@ -18,13 +20,21 @@ def reconstruct(file_pk):
...
             ext = f'.{ref}.emlxpart'
             part_name = re.sub(r'\.partial\.emlx$', ext, file.name)
+            parent = file.parent_directory
+
+            require_dependency(
+                f'walk-{parent.pk}', depends_on,
+                lambda: filesystem.walk.laterz(parent.pk),
+            )
+
+            require_dependency(
```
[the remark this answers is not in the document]
mgax: Yep, good catch! I think the coorect solution is to mark the `walk_file` task as `broken` and `emlx.reconstruct` should handle it gracefully.

PR #37, opened 2018-02-19
only write new blobs to the storage

gabriel-v:
With this patch, walk doesn't overwrite existing blobs in the blob storage.

### DIFF
--- a/snoop/data/models.py
+++ b/snoop/data/models.py
@@ -6,6 +6,7 @@ from django.db import models
 from django.conf import settings
 from django.template.defaultfilters import truncatechars
 from django.contrib.postgres.fields import JSONField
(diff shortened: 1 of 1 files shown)


snoop/data/models.py
```diff
@@ -117,19 +118,35 @@ def create(cls):
...
-        with cls.create() as writer:
-            writer.write(data)
+        sha3_256 = hashlib.sha3_256()
+        sha3_256.update(data)
 
-        return writer.blob
+        try:
+            return Blob.objects.get(pk=sha3_256.hexdigest())
+
+        except ObjectDoesNotExist:
```
mgax: You can do `except Blob.DoesNotExist` afair

PR #42, opened 2018-02-21
parse exif data

mgax:


snoop/data/analyzers/exif.py
```diff
@@ -0,0 +1,71 @@
...
+    data = {}
+    gps = extract_gps_location(tags)
+    if gps:
+        data['location'] = gps
+
+    for key in ['EXIF DateTimeOriginal', 'Image DateTime']:
+        if key in tags:
+            date = convert_exif_date(str(tags[key]))
+            if date:
+                data['date-created'] = date
```
gabriel-v: I think we want this date in the `Z` format by using `utlis.zulu` or something. Also see the `convert_exif_date` function above.

mgax: Yeah, makes sense, done. And we'll quietly ignore the fact that exifs don't have timezone information and we just assume UTC.

PR #44, opened 2018-02-28
command to retry tasks by func and status

mgax:


snoop/data/management/commands/retrytasks.py
```diff
@@ -0,0 +1,33 @@
...
+        if func:
+            queryset = queryset.filter(func=func)
+
+        status = options.get('status')
+        if status:
+            queryset = queryset.filter(status=status)
+
+        if options.get('n'):
+            print(queryset.count())
+
```
gabriel-v: why the whitespace?

mgax: I always leave an empty line after an indent block, I'm surprised you're bringing it up :)

PR #61, opened 2018-03-13
Date fields

mgax:
closes https://github.com/hoover/snoop2/issues/45

### DIFF
--- a/snoop/data/analyzers/tika.py
+++ b/snoop/data/analyzers/tika.py
@@ -5,6 +5,7 @@ import requests
 from dateutil import parser
 from ..tasks import shaorma, ShaormaBroken, returns_json_blob
 from .. import models
+from ..utils import zulu
 
 TIKA_CONTENT_TYPES = [
     'text/plain',
@@ -81,7 +82,7 @@ def get_date_created(rmeta):
     for field in FIELDS_CREATED:
         value = rmeta[0].get(field)
         if value:
-            return parser.parse(value).isoformat()
+            return zulu(parser.parse(value))
 
 def get_date_modified(rmeta):
     FIELDS_MODIFIED = ['Last-Modified', 'Last-Saved-Date', 'dcterms:modified',
@@ -90,4 +91,4 @@ def get_date_modified(rmeta):
     for field in FIELDS_MODIFIED:
         value = rmeta[0].get(field)
         if value:
-            return parser.parse(value).isoformat()
+            return zulu(parser.parse(value))
--- a/snoop/data/analyzers/tika.py
+++ b/snoop/data/analyzers/tika.py
@@ -2,6 +2,7 @@ from urllib.parse import urljoin
 import json
 from django.conf import settings
 import requests
+from dateutil import parser
 from ..tasks import shaorma, ShaormaBroken, returns_json_blob
 from .. import models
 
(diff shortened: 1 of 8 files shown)


snoop/data/analyzers/tika.py
```diff
@@ -71,3 +72,22 @@ def rmeta(blob):
...
+
+
+def get_date_created(rmeta):
+    FIELDS_CREATED = ['Creation-Date', 'dcterms:created', 'meta:created',
+                      'created']
+
+    for field in FIELDS_CREATED:
+        value = rmeta[0].get(field)
+        if value:
+            return parser.parse(value).isoformat()
```
gabriel-v: `utils.zulu` format?

snoop/data/analyzers/tika.py
```diff
@@ -71,3 +72,22 @@ def rmeta(blob):
...
+
+def get_date_modified(rmeta):
+    FIELDS_MODIFIED = ['Last-Modified', 'Last-Saved-Date', 'dcterms:modified',
+                       'meta:modified', 'created']
+
+    for field in FIELDS_MODIFIED:
+        value = rmeta[0].get(field)
+        if value:
+            return parser.parse(value).isoformat()
```
gabriel-v: `utils.zulu` format?

snoop/data/digests.py
```diff
@@ -134,12 +136,18 @@ def iter_parts(email_data):
     for header in ['To', 'Cc', 'Bcc', 'Resent-To', 'Recent-Cc']:
         headers_to.update(headers.get(header, []))
 
+    message_date = None
+    message_raw_date = headers.get('Date', [None])[0]
+    if message_raw_date:
+        message_date = email.parse_date(message_raw_date).isoformat()
```
gabriel-v: `utils.zulu` format?

mgax:
> Why not use the utils.zulu format

Yeah good catch.

gabriel-v:
You missed a few `+00:00` in your tests I think.

But we have another problem: that `zulu` function does assert `date.endswith('+00:00')`. What if we get a date with a specific timezone from tika or the email file?

Can we fix & test the `zulu` function so if the date given to it has a timezone, it gets changed to UTC?

mgax:
Hmm, `zulu` only works for timestaps in the UTC timezone. And we don't have to use it, any iso-formatted date should work fine with elasticsearch. Reverting the last patch.

gabriel-v:
Then why are we using `zulu` anywhere?

mgax:
Can't remember. Was it because ES didn't always auto-detect them as dates?

mgax:
There: https://github.com/hoover/snoop2/blob/52665c93208a96c3ed968b2582a4821a5b3fbbbd/snoop/data/utils.py#L33

gabriel-v:
Looking at the docs, yeah, timezones should be supported.

Do run some checks about it, and look at https://www.elastic.co/guide/en/elasticsearch/reference/2.4/date.html which says they turn all dates into UTC anyway. So we might as well be safe and convert them to the UTC Z format as they want it.

gabriel-v:
travis: `NameError("name 'zulu' is not defined",)` in the integration test

mgax:
I made some tests, ES indexes them properly, and filtering by date/time range works.

PR #62, opened 2018-03-13
Handle broken dependencies

mgax:
Fixes https://github.com/hoover/snoop2/issues/39

### DIFF
--- a/snoop/data/digests.py
+++ b/snoop/data/digests.py
@@ -43,7 +43,7 @@ def gather(blob, collection_pk, **depends_on):
     if tika_rmeta_blob:
         if isinstance(tika_rmeta_blob, ShaormaBroken):
             rv['broken'].append(tika_rmeta_blob.reason)
-            log.warn("tika_rmeta task is broken; skipping")
+            log.debug("tika_rmeta task is broken; skipping")
 
         else:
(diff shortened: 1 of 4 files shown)


snoop/data/digests.py
```diff
@@ -38,15 +41,25 @@ def gather(blob, collection_pk, **depends_on):
 
     tika_rmeta_blob = depends_on.get('tika_rmeta')
     if tika_rmeta_blob:
-        with tika_rmeta_blob.open(encoding='utf8') as f:
-            tika_rmeta = json.load(f)
-        rv['text'] = tika_rmeta[0].get('X-TIKA:content', "")
+        if isinstance(tika_rmeta_blob, ShaormaBroken):
+            rv['broken'].append(tika_rmeta_blob.reason)
+            log.warn("tika_rmeta task is broken; skipping")
```
gabriel-v: This means a lot of log traffic. Don't think it's of any use, you should see the `broken` list in the digest, right?

mgax: Right. How about I change those to `log.debug`?

gabriel-v: sure

gabriel-v: Merge it afterwards

PR #65, opened 2018-03-15
Index data using the digests.index task

gabriel-v:


.travis.yml
```diff
@@ -1,32 +1,15 @@
...
-
-env:
-  - SNOOP_TESTDATA=testdata
-
-before_script:
-  - java -jar tika-server-1.17.jar &
-  - sleep 5
+  - docker build . --tag snoop2-testing
+  - git clone https://github.com/hoover/docker-setup.git
+  - ( cd docker-setup && git checkout snoop2-gather-index-into-es && cp docker-compose.override.travis-snoop2.yml docker-compose.override.yml && cp snoop.testing.env snoop.env && cp search.testing.env search.env && cp nginx.testing.env nginx.env && docker-compose up -d )
```
mgax: I'm ok with having `*.testing.env` in the `docker-setup` repo, but please move them to a subfolder, so they don't spam the root folder. `docker-compose.override.travis-snoop2.yml` should be in the `snoop2` repo though.

Also, before we merge this, let's merge the `snoop2` branch of the `docker-setup` repo. Here is a PR: https://github.com/hoover/docker-setup/pull/3

gabriel-v:
Don't merge this until https://github.com/hoover/docker-setup/pull/4 is merged and the line `git checkout snoop2-gather-index-into-es` is removed from `.travis.yml`

mgax:
`addcollection` should not clobber your elasticsearch index without warning. Either be explicit about it, or somehow auto-detect that the index is missing and create it.

snoop/data/digests.py
```diff
@@ -87,13 +88,27 @@ def gather(blob, collection_pk, **depends_on):
...
 
-    collection.digest_set.update_or_create(
+    digest, _ = collection.digest_set.update_or_create(
         blob=blob,
         defaults=dict(
             result=writer.blob,
         ),
     )
 
+    index.laterz(digest.pk)
```
mgax: How can we create these tasks for existing digests? So we can reuse the digest blobs that have already been created.

snoop/data/management/commands/createcollection.py
```diff
@@ -1,5 +1,7 @@
 from pathlib import Path
+import requests
```
mgax: No need for requests here.

snoop/defaultsettings.py
```diff
@@ -62,9 +62,11 @@
 STATIC_URL = '/static/'
 STATIC_ROOT = str(base_dir / 'static')
 
+SNOOP_COLLECTION_ELASTICSEARCH = None
```
mgax: It feels like plural works better - `SNOOP_COLLECTIONS_ELASTICSEARCH`

snoop/data/digests.py
gabriel-v: Rerunning all `digests.gather` tasks?

mgax: Yeah, but that's i/o intensive, they each read a bunch of blobs, and write the output blob.

gabriel-v: Then I'll write you a one-liner to pass to django shell.

mgax: Perfect, thanks!

PR #73, opened 2018-04-10
only select files from the same collection

mgax:


snoop/data/digests.py
```diff
@@ -189,7 +189,13 @@ def iter_parts(email_data):
 
 
 def _get_document_content(digest):
-    first_file = digest.blob.file_set.order_by('pk').first()
+    first_file = (
```
gabriel-v: Now that you've changed this, it looks more like a method on `models.Digest` if you ask me.

mgax: Yeah, agreed, but I'll merge it as is to fix the bug.

PR #76, opened 2018-04-11
Unpack mbox files

mgax:


snoop/data/analyzers/archives.py
```diff
@@ -59,6 +69,33 @@ def call_7z(archive_path, output_dir):
...
+def unpack_mbox(mbox_path, output_dir):
+    def slice(stream):
+        last = b''
+        while True:
+            buffer = stream.read(1024 * 64)
+            if not buffer:
+                break
+            window = last + buffer
+            while True:
+                m = re.search(br'\n\r?\n(From )', window)
```
gabriel-v: Use case insensitive search here please.

snoop/data/analyzers/archives.py
```diff
@@ -59,6 +69,33 @@ def call_7z(archive_path, output_dir):
...
+                    break
+                offset = m.start(1)
+                yield window[:offset]
+                window = window[offset:]
+            last = window
+        yield last
+
+    with open(mbox_path, 'rb') as f:
+        for n, message in enumerate(slice(f), 1):
+            hash = sha1(str(n).encode('utf-8')).hexdigest()
```
gabriel-v: Why'd we want to hash the email index `n`? Can't we use it as it is?

snoop/data/analyzers/archives.py
mgax: No, it's good.

snoop/data/analyzers/archives.py
mgax: Yeah i just copy pasted the code from https://github.com/hoover/scripts/blob/master/mbox_to_eml.py.

PR #85, opened 2018-05-07
Binary filenames

### LINKED ISSUE #81: Store filenames as bytes
In the case of broken filenames (bytes that are not valid utf-8) snoop is not able to write the filenames to the database. We could store the filenames as raw bytes and as a cleaned-up text version for display purposes.

mgax:
This patch changes the `name` field of `Directory` and `File` to a `name_bytes` field of type `BinaryField` so it can store surrogateescape-encoded values (see https://docs.python.org/3/library/codecs.html#error-handlers). It also changes all reads, writes and filtering to those fields.

The test suite passes and a manual check shows that snoop can process files with broken (non-utf8) filenames under Linux.

Fixes #81


testsuite/test_emails.py
```diff
@@ -194,6 +194,7 @@ def test_emlx_reconstruction_with_missing_dependency(taskmanager):
     d3 = mkdir(d2, 'Data')
     d4 = mkdir(d3, '1')
     d5 = mkdir(d4, 'Messages')
+
```
gabriel-v: :clap:

PR #92, opened 2018-06-08
 Import from tarballs using reading with transparent compression.

salevajo:


salevajo:
As the command line usage worked for all commands, I checked the test case itself and realised, that we need to open the test files using mode 'w+b' with the additional '+' for updating (reading and writing). Local test are now working again.

PR #101, opened 2018-07-19
Update index mapping for ES6

jarib:
See hoover/docker-setup#11


jarib:
The tests are passing locally, but failing on Travis. 

I made an attempt to make Travis use the `es6` branch of hoover/docker-setup if this PR is the one being built, but something is not working correctly with the Docker networking, which I'm not so familiar with. The result is that `http://search-es:9200` cannot be reached from the test suite.

I'll look into it more tomorrow. Suggestions welcome!

ioanvpocol:
After the docket-setup PR is merged I think you can delete the TRAVIS_PULL_REQUEST_BRANCH related code.

jarib:
Yes, that’s the plan!

It seems the Travis trouble I was having was simply related to the ES startup time. Not sure why it’s longer than it used to be. 

I’ll try to implement a solution for waiting that doesn’t use `sleep` and clean up the Travis config, then this should be ok to merge.

ioanvpocol:
Maybe you can uset this:  https://github.com/vishnubob/wait-for-it
It is used also on docker file:  https://github.com/hoover/snoop2/blob/master/Dockerfile#L45

.travis.yml
```diff
@@ -4,17 +4,16 @@ services:
...
   - docker build .. --tag snoop2-testing
   - cp ../testsuite/docker-compose.override.travis-snoop2.yml docker-compose.override.yml
   - cp snoop.testing.env snoop.env
   - cp search.testing.env search.env
   - docker-compose up -d
   - git clone https://github.com/hoover/testdata collections/testdata
-
 script:
-  - docker-compose run --rm snoop bash -c 'sleep 15 && py.test -v'
+  - docker-compose run --rm snoop bash -c '/wait && py.test -v'
```
ioanvpocol: Is wait accessible here?

ioanvpocol:
:+1:

PR #132, opened 2019-01-10
fix certain indexing errors

mugurrus:


snoop/data/digests.py
```diff
@@ -102,7 +102,12 @@ def index(blob, digests_gather):
     content = _get_document_content(digest)
     version = _get_document_version(digest)
     body = dict(content, _hoover={'version': version})
-    indexing.index(digest.blob.pk, body)
+    try:
+        indexing.index(digest.blob.pk, body)
+    except RuntimeError as e:
+        log.exception(repr(e))
+        log.exception(repr(body))
```
mgax: `log.exception` already prints the error and stack trace so the 2nd line is enough.

snoop/data/digests.py
```diff
@@ -102,7 +102,12 @@ def index(blob, digests_gather):
...
+        raise e
```
mgax: Or just write `raise`, it does the same thing.

snoop/data/indexing.py
```diff
@@ -85,8 +86,11 @@ def check_response(resp):
 
 
 def index(id, data):
-    if settings.DETECT_LANGUAGE and data.get('text', ''):
-        data['lang'] = langdetect.detect(data.get('text', '')[:2500])
+    if settings.DETECT_LANGUAGE and data.get('text') is not None:
+        try:
+            data['lang'] = langdetect.detect(data['text'][:2500])
+        except LangDetectException:
+            data['lang'] = None
```
mgax: Should we not log this error, at least with log level `DEBUG`?

snoop/data/tasks.py
```diff
@@ -183,6 +184,15 @@ def run_task(task_pk, raise_exceptions=False):
                 log=handler.stream.getvalue(),
             )
 
+        except ConnectionError as e:
+            logger.exception(repr(e))
```
mgax: Again, `logger.exception` prints all the exception details, you should supply a human-readable message explaining when the error occurred.

PR #136, opened 2019-02-03
Liquid nomad

mgax:
Improve the docker configuration so that it can pull values directly fom the environment without the need to mount a config file.


snoop/wsgi.py
```diff
@@ -9,5 +9,6 @@
 
 from django.conf import settings
 from . import defaultsettings
-if not settings.DEBUG and settings.SECRET_KEY == defaultsettings.SECRET_KEY:
-    raise RuntimeError("Please change the default SECRET_KEY setting")
+if settings.SECRET_KEY == defaultsettings._default_secret_key:
```
mugurrus: Why split in 2 branches?

mgax: To avoid a long line, so that it's easier to read.

mugurrus: To me it was perfectly readable as it was. Adding more branches increases complexity, while consolidating them reduces it.

mgax: Ok, how about now?

mugurrus: OK

mgax:
@mugurrus Do you think this could break `docker-setup`? If not, I will merge.

PR #137, opened 2019-02-14
optimize blob writing

mugurrus:


snoop/data/models.py
```diff
@@ -201,7 +203,7 @@ def name(self):
     def parent(self):
         return self.parent_directory or self.container_file
 
-    def ancestry(item):
+    def ancestry(self, item):
```
mgax: Are you sure you want to add `self` here?

mgax: The code, as it was, functioned correctly. `item` is initially the model instance, and then changes to its parent, and so on. I guess it would have been more clear like so:

```python
def ancestry(self):
    item = self
    while item:
        yield item
        item = item.parent
```

PR #141, opened 2019-03-08
Skip the child dirs of container files when moving up the parent chain

gabriel-v:
Fixes https://github.com/CRJI/EIC/issues/110

Still needs some tests.

### DIFF
--- a/snoop/data/digests.py
+++ b/snoop/data/digests.py
@@ -154,6 +154,9 @@ def parent_id(file):
         return parent.blob.pk
 
     if isinstance(parent, models.Directory):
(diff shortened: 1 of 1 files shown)


gabriel-v:
Initially we talked about changing [this line](https://github.com/hoover/snoop2/blob/0dc232baf91be7b6c878e095a2939f696be665d2/snoop/data/digests.py#L294) to stop skipping over the single child directory of container files (emails, archives).

I think it's simpler if we skip those dirs like we've been doing until now. Fixing this ticket means making the `parent_id` keys skip them too, not just the `children` list.

PR #145, opened 2019-04-02
Auto-sync feature for collections

MattesR:
periodically runs a walk task on the filesystem to sync the files in the collection with hoover


snoop/data/management/commands/runworkers.py
```diff
@@ -54,11 +54,13 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument('func', nargs='*',
-                help="Task types to run")
+                            help="Task types to run")
```
mgax: Please fix whitespace in a separate PR.

PR #150, opened 2019-04-29
changed sync to work via environment variable

MattesR:


snoop/data/management/commands/runworkers.py
```diff
@@ -72,7 +72,7 @@ def handle(self, *args, **options):
                 prefix = settings.TASK_PREFIX
             queues = options.get('func') or tasks.shaormerie
             system_queues = ['watchdog']
-            if options.get('synced'):
+            if bool_env(os.environ.get('SYNCED')):
```
mgax: Let's call the env var `SYNC_FILES`. By the way, "sync" is preferrable to "synced", because it describes the action that snoop should take.

MattesR:
I'll add a unit test and I put it in /testsuites.

snoop/data/management/commands/runworkers.py
MattesR: Okay, I can see that.

MattesR:
Unit test is not written yet, but as we discussed in the meeting, I've created an Issue and we could already merge.

PR #153, opened 2019-05-13
Updates

MattesR:
I've briefly summarized the things I did in this Gist:

https://gist.github.com/MattesR/a8adca10a5d52f33723218e049b2ecb3

Since I don't really know how Whitenoise works, I can't asses whether it still does what it's supposed to do, but I at least you can retrace my steps.


gabriel-v:
@MattesR Whitenoise serves static files. You can do a `GET /static/something.css

requirements.txt
```diff
@@ -1,62 +1,63 @@
...
-autopep8==1.3.5           # via django-silk
-babel==2.6.0              # via flower
-billiard==3.5.0.3         # via celery
-bleach==2.1.3
-cachetools==3.1.0         # via google-auth
-celery==4.1.0
-certifi==2018.4.16        # via requests
-chardet==3.0.4            # via requests
-click==6.7                # via pip-tools
+-i https://pypi.org/simple
```
mgax: You can remove `requirements.txt` if we have a pipfile. And let's replace `pip install -r requirements.txt` from the [Dockerfile](https://github.com/hoover/snoop2/blob/96d1e72b2cece28fdd668f784d9c1c0131aaf47e/Dockerfile#L23) with `pipenv install`:

https://github.com/hoover/reup/blob/37c947ec8f51db3b36f9cf5f8313b269c7acc06b/Dockerfile#L10-L12

snoop/trace.py
```diff
@@ -7,11 +7,11 @@
 from opencensus.stats import measure as measure_module
 from opencensus.stats import stats as stats_module
 from opencensus.stats import view as view_module
-from opencensus.stats.exporters import prometheus_exporter as prometheus
+from opencensus.ext.prometheus import stats_exporter as prometheus
```
mgax: Can we remove the prometheus thing? AFAIK we're not using it, and it's causing trouble on nomad. @gabriel-v, @mugurrus, thoughts?

snoop/defaultsettings.py
```diff
@@ -34,6 +34,7 @@
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    'whitenoise.middleware.WhiteNoiseMiddleware',
```
mgax: :+1:

snoop/trace.py
gabriel-v: I think the update is good, since it's just following the changelog. I vote to fix the whole tracing situation in another pull request and leave this patch as it is.

MattesR:
>     * add `Pipfile.lock` to git because it's omitted

Isn't `Pipfile.lock` generate by pipenv when creating the environment? I  deliberately omitted it because I thought that generated files shouldn't be on version control.

mgax:
> Isn't `Pipfile.lock` generate by pipenv when creating the environment? I deliberately omitted it because I thought that generated files shouldn't be on version control.

They [recommend](https://docs.pipenv.org/en/latest/basics/#general-recommendations-version-control) to keep it versioned.

mgax:
@MattesR please fix the conflicts and merge

PR #167, opened 2019-06-20
Use `gpghome` folder which is already set up

mgax:
The PGP code was always kind of a hack, and it expects a gnupg home folder to be set up. So https://github.com/hoover/testdata/commit/c7c3bee47b431c43d4bbdfa1bae720e2dfd0e6d8 creates that folder in testdata, next to `data` and `ocr`. The idea is that, if you want pgp snooping, you set up the home folder in the collection folder, and snoop will auto-detect it.

### DIFF
--- a/snoop/data/analyzers/pgp.py
+++ b/snoop/data/analyzers/pgp.py
@@ -1,6 +1,9 @@
+from pathlib import Path
 import subprocess
 from django.conf import settings
 from ..tasks import ShaormaBroken
+
+gpghome = Path(settings.SNOOP_COLLECTION_ROOT).parent / 'gpghome'
 
 
 def is_encrypted(data):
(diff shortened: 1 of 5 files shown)


snoop/data/analyzers/pgp.py
```diff
@@ -1,18 +1,21 @@
+from pathlib import Path
 import subprocess
 from django.conf import settings
 from ..tasks import ShaormaBroken
 
+gpghome = Path(settings.SNOOP_COLLECTION_ROOT).parent / 'gpghome'
```
mugurrus: The default gpg home dir is `gnupg` (see https://github.com/liquidinvestigations/node/blob/master/templates/collection.nomad#L201). This is related to https://github.com/liquidinvestigations/node/pull/116

mgax: Oh, I totally missed that line, thanks! But I still think `gpghome` should be in the collection folder, it just makes more sense.

testsuite/test_integration.py
```diff
@@ -76,8 +76,10 @@ def feed_page(url):
     assert es_count == db_count
 
     # check that all index ops were successful
-    db_failed_count = models.Task.objects.filter(func='digests.index').exclude(status='success').count()
-    assert db_failed_count == 0
+    index_failed = [(t.args, t.status) for t in models.Task.objects.filter(func='digests.index').exclude(status='success')]
+    # one indexing task should be deferred because
+    # `encrypted-hushmail-smashed-bytes.eml` is broken
+    assert index_failed == [(['66a3a6bb9b8d86b7ce2be5e9f3a794a778a85fb58b8550a54b7e2821d602e1f1'], 'deferred')]
```
gabriel-v: wunderbar

PR #170, opened 2019-07-02
Enable flake8 check

### LINKED ISSUE #169: Enable flake8 check
https://github.com/hoover/snoop2/blob/dbdd60aced79ec50f0ecce93a4619d50d5b54211/.drone.yml#L9

raduklb:
Enabled flake8 check and fixed all coding style issues so flake8 would pass.

Closes #169


snoop/trace.py
```diff
@@ -1,17 +1,15 @@
...
-from opencensus.stats import aggregation as aggregation_module
-from opencensus.stats import measure as measure_module
-from opencensus.stats import stats as stats_module
-from opencensus.stats import view as view_module
-from opencensus.ext.prometheus import stats_exporter as prometheus
-from opencensus.tags import tag_key as tag_key_module
-from opencensus.tags import tag_map as tag_map_module
-from opencensus.tags import tag_value as tag_value_module
-from opencensus.ext.zipkin import trace_exporter as zipkin
+from opencensus.stats import aggregation as aggregation_module  # noqa: F401
```
gabriel-v: Why import these if they're unused? Who knows where the copy/pasta came about.

testsuite/test_blobs.py
```diff
@@ -1,11 +1,11 @@
-from pathlib import Path
+# flake8: noqa
```
gabriel-v: I think we should have another `.flake8` under `testsuite` that excludes more rules, like:

- max line lengths
- star imports

And remove this file ignore.

testsuite/test_emails.py
```diff
@@ -15,9 +15,12 @@
 DATA = Path(settings.SNOOP_TESTDATA) / "data"
 EML = DATA / "eml-5-long-names/Attachments have long file names..eml"
 MSG = DATA / "msg-5-outlook/DISEARĂ-Te-așteptăm-la-discuția-despre-finanțarea-culturii.msg"
-MAPBOX = DATA / "eml-1-promotional/Introducing Mapbox Android Services - Mapbox Team <newsletter@mapbox.com> - 2016-04-20 1603.eml"
-CODINGAME = DATA / "eml-1-promotional/New on CodinGame: Check it out! - CodinGame <coders@codingame.com> - 2016-04-21 1034.eml"
-CAMPUS = DATA / "eml-2-attachment/FW: Invitation Fontys Open Day 2nd of February 2014 - Campus Venlo <campusvenlo@fontys.nl> - 2013-12-16 1700.eml"
+MAPBOX = DATA / ("eml-1-promotional/Introducing Mapbox Android Services - Mapbox Team "
```
gabriel-v: Same as above: if the line length has no limit, let's not line break here.

.drone.yml
```diff
@@ -3,11 +3,9 @@ name: test
 
 steps:
 - name: flake8
-  image: python:3.7-stretch
+  image: alpine/flake8:3.7.7
   commands:
-  - pip install flake8 > /dev/null
-  - flake8 || true
-  - echo 'not running flake8 yet'
+  - flake8 --ignore=W503
```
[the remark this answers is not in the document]
gabriel-v: Seems [per-directory-config](https://github.com/PyCQA/pylint/projects/1) isn't a thing yet. Sorry about that. We should only have one .flake8 file in the project root, probably styled like this: https://flake8.pycqa.org/en/latest/user/configuration.html

snoop/data/migrations/0001_initial.py
```diff
@@ -14,7 +14,8 @@ class Migration(migrations.Migration):
         migrations.CreateModel(
             name='Collection',
             fields=[
-                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False,
```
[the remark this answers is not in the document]
gabriel-v: > so please add a `.flake8` file to `snoop/data/migrations/` 

I was mistaken, there should only be one `.flake8` file in the root. I think it's best to exclude the `migrations` folder completely from that file.

.drone.yml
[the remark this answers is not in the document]
raduklb: From the first link you pointed regarding W503: 

> This rule goes against the PEP 8 recommended style, which was changed on April 16th, 2016 in this [commit](https://github.com/python/peps/commit/c59c4376ad233a62ca4b3a6060c81368bd21e85b).

If I would fix W503, then a W504 error would show up -- W504 is the rule to be followed.

.flake8
```diff
@@ -0,0 +1,7 @@
+[flake8]
+ignore = W503
+exclude = 
+    snoop/data/migrations/,
+    snoop/wsgi.py	  
+per-file-ignores =
```
gabriel-v: wonderful

snoop/data/magic.py
```diff
@@ -103,7 +103,7 @@ def download_magic_definitions():
     which_file = subprocess.check_output(['which', 'file']).decode('latin1')
     version_file = subprocess.check_output([
         'file', '--version',
-         '-m', str(MAGIC_FILE),
```
gabriel-v: I expected more of these

.flake8
```diff
@@ -0,0 +1,7 @@
+[flake8]
+ignore = W503
+exclude = 
+    snoop/data/migrations/,
```
gabriel-v: Instead of excluding the migrations, could we put this under `per-test-ignores`? (just need to ignore max line length I think)

Once every half a year we end up writing migrations by hand and it would be really useful to have this check here.

PR #188, opened 2019-08-27
implement entity detection functionality

mugurrus:


gabriel-v:
to be continued after https://github.com/CRJI/EIC/issues/331

PR #202, opened 2019-09-18
Decrease load impact from sync

gabriel-v:
We could add some `sleep 100ms` instructions in between the auto-sync job retries to spread their runtime over a longer period, hopefully decreasing load.

Any ideas?


mgax:
So this means we auto-sync every 15 minutes now? @germakoci is that ok?

germakoci:
this means that uploading a doc in nextcloud will take 15 min to become searchable in hoover ? if that is the question, yes, totally acceptable, if this gives us more speed

gabriel-v:
We're probably not getting speed from it, just a lower average load on the server. 

I'll set it at 7 minutes, it should make a difference with that interval too.

PR #220, opened 2019-10-28
update repo urls

gabriel-v:


factory-scripts/guest/setup-artful.sh
```diff
@@ -41,10 +41,10 @@ EOF
 supervisorctl update
 
 # download testdata
-git clone https://github.com/hoover/testdata.git /opt/testdata
+git clone https://github.com/liquidinvestigations/testdata.git /opt/testdata
```
mgax: Let's call it `hoover-testdata`, it's intimately related to hoover.

PR #228, opened 2019-11-01
Connection backlog

gabriel-v:
Fix an issue with directory queries not being limited in result size.

We should open an issue about paginating these views.


Dockerfile
```diff
@@ -45,7 +45,7 @@ ADD https://github.com/ufoscout/docker-compose-wait/releases/download/2.3.0/wait
 
 RUN set -e \
  && echo '#!/bin/bash -e' > /runserver \
- && echo 'waitress-serve --port 80 snoop.wsgi:application' >> /runserver \
+ && echo 'waitress-serve --threads 42 --port 80 snoop.wsgi:application' >> /runserver \
```
mgax: I really want this to be configurable using an environment variable. In case we want to change it on a broken production system without having to rebuild the whole docker container, on possibly failed CI infrastructure ;)

PR #264, opened 2020-01-10
Log worker events

gabriel-v:
Rearrange some logging messages, fix `%r` of a `snoop.data.models.Task` being `''` in the current configuration, and add one `.iterator()` call where it should've been.

### DIFF
--- a/snoop/data/models.py
+++ b/snoop/data/models.py
@@ -78,6 +78,8 @@ class Blob(models.Model):
     def __str__(self):
         return self.pk
 
+    __repr__ = __str__
+
     @property
     def content_type(self):
         if self.mime_type.startswith('text/'):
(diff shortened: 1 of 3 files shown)


snoop/data/models.py
```diff
@@ -78,6 +78,8 @@ class Blob(models.Model):
     def __str__(self):
         return self.pk
 
+    __repr__ = __str__
+
```
mgax: I'm pretty sure this should not be necessary, the default `__repr__` is sane ([on master](https://github.com/django/django/blob/5a4d7285bd10bd40d9f7e574a7c421eb21094858/django/db/models/base.py#L518-L519); also [in version 2.2.8](https://github.com/django/django/blob/2.2.8/django/db/models/base.py#L518-L519) that we currently use)

snoop/data/tasks.py
```diff
@@ -284,21 +283,30 @@ def dispatch_pending_tasks():
         ])
     )
 
-    for task in task_query:
+    task_count = task_query.count()
```
mgax: In Postgres this is [surprisingly expensive](https://wiki.postgresql.org/wiki/Slow_Counting) if there are many rows in the count. It might be a better idea to query for a single item and see if you receive anything.

PR #284, opened 2020-02-23
Check if work is done

mgax:
Tested manually, works as a charm.


snoop/data/management/commands/workisdone.py
```diff
@@ -0,0 +1,24 @@
...
+
+        queryset = (
+            models.Task.objects
+            .filter(status__in=[
+                models.Task.STATUS_PENDING,
+                models.Task.STATUS_DEFERRED,
+            ])
+        )
+
+        if queryset.first() is not None:
```
gabriel-v: `if queryset.exists():`

Issue #159, opened 2019-06-06
Add view for the File model


gabriel-v:
We have views for Digests and Directories, but not Files.

The UI needs access to the File model to properly show the file tree structure, since the digests are de-duplicated and will contain bad parent links.

The File view should probably also return the digest for that file.

mgax:
That sounds reasonable. It would make sense to expose this structure with graphql.

gabriel-v:
I wasn't suggesting we add graphql with this, just a new endpoint at https://github.com/hoover/snoop2/blob/dbdd60aced79ec50f0ecce93a4619d50d5b54211/snoop/data/views.py#L43 that gets the data for a File. This endpoint would be used in the UI file tree view.

PR #331, opened 2020-08-24
Add view and API entry for File model

RazorBest:
I added a view for models.File. It is similar to the Directory and Digest views.
Now, every file can be accessed with a unique id in the format "\_file\_\<pk\>", instead of a sha3 of its contents.

Solves #159

### DIFF
--- a/snoop/data/digests.py
+++ b/snoop/data/digests.py
@@ -186,11 +186,15 @@ def directory_id(directory):
     return f'_directory_{directory.pk}'
 
 
+def file_id(file_query):
+    return f'_file_{file_query.pk}'
+
+
 def parent_id(file):
     parent = file.parent
 
     if isinstance(parent, models.File):
-        return parent.blob.pk
+        return file_id(parent)
 
     if isinstance(parent, models.Directory):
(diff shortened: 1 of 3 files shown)


gabriel-v (changes requested):
I think it would make sense to integrate these file IDs into the rest of the system:

- in the `parent_id` function here:
```
def parent_id(file):
    parent = file.parent
    if isinstance(parent, models.File):
        return parent.blob.pk
    ...
```
We should return the file id, not the digest.

- in the `get_document_locations` function here:
```
def get_document_locations(digest):
    def location(file):
        parent_path = full_path(file.parent_directory.container_file or file.parent_directory)
        return {
            'filename': file.name,
            'parent_id': parent_id(file),
            'parent_path': parent_path,
        }
    ...
```

we should add here alongside `filename` the `id` as `file_id(file)`


We need to make sure any changes made here won't break the UI, I expect this to happen in places where the UI expects to get a hash in the `id` slot of some object, and it will get a `_file_500` id instead. Feel free to open another PR in the hoover-ui repository and fix any such breakage.

RazorBest:
Move from fork to main repo

Issue #83, opened 2018-04-24
Change ocr.walk_file pattern matching


mgax:
ocr.walk_file should look at the first 32 chars of the filename and try to match it as an md5 hash. The current code also concatenates parent folder names; this is not needed.

Relevant line: https://github.com/hoover/snoop2/blob/1864656db28f0c9b81c06f2a4e74acb264cc011a/snoop/data/ocr.py#L60

PR #333, opened 2020-09-02
File model view

RazorBest:
I added a view for models.File. It is similar to the Directory and Digest views.
Now, every file can be accessed with a unique id in the format "\_file\_\<pk\>", instead of a sha3 of its contents.

Solves #159

<details>
<summary> Here's an example of what the API returns for a directory: </summary>

Before:
```yaml
$ curl -L http://10.66.60.1:9990/snoop/collections/testdata/_directory_15/json

{
    "id": "_directory_15",

    ...

    "children": [{
        "id": "c0479bc7a3d445cab5e84c9e69275a77179f24562c727dd034b32e364c338c6d",
        "content_type": "message/rfc822",
        "filename": "broken-subject.eml"
    }],

    ...
}
```

After:
```yaml
$ curl -L http://10.66.60.1:9990/snoop/collections/testdata/_directory_15/json

{
    "id": "_directory_15",

    ...

    "children": [{
        
        "id": "_file_7",
        "digest": "c0479bc7a3d445cab5e84c9e69275a77179f24562c727dd034b32e364c338c6d",
        "content_type": "message/rfc822",
        "filename": "broken-subject.eml"
    }],

    ...
}
```
</details>

As we can see, before, the sha3 was accessed through the "id" field. Now, it can be accessed through the "digest" field.
It's clear that when we only need the contents of the file, we do it through the "digest" field. But when we need other metadata (eg. absolute path), we should do it through the "id" field - and this hasn't been implemented yet on the client side.

There are still parts of the code that access the sha3 through the "id" field and only need the content of the file (eg. in hoover-ui). So, if this merge takes place, we should switch from "id" to "digest" in this cases, on the client side.

I think there are also places that don't need just the content of the file. They also need some metadata, like mentioned earlier,  but they still access the file through the sha3. Here, I think we should restructure how the client calls the API. They should rely on the File view instead of the Digest view. It think this approach should solve https://github.com/CRJI/EIC/issues/110


gabriel-v (approved):
This LGTM but we can't merge it since the UI will break, as you said.

Please continue work by duplicating [this branch](https://github.com/liquidinvestigations/hoover-ui/tree/enable-tree-view), rebasing it, and fixing references from `.id` to `.digest` when navigating a File instance. Please leave the 2019 branch unchanged for reference.

Continue that work as a part of https://github.com/CRJI/EIC/issues/83 and open a new pull request in the `hoover-ui` repo with these changes.

RazorBest:
Technically, this solves https://github.com/CRJI/EIC/issues/110 and doesn't break the UI. But that's only because the File API replicates the Digest API.

snoop/data/views.py
```diff
@@ -61,6 +61,50 @@ def feed(request):
...
 
 
+@collection_view
+def file_view(request, pk):
+    file = get_object_or_404(models.File.objects, pk=pk)
+    return JsonResponse(trim_text(digests.get_file_data(file)))
+
+
+@collection_view
+def file_download(request, pk, filename):
```
gabriel-v, line 71: please remove the following 3 functions

gabriel-v (changes requested):

gabriel-v (approved):
we'll merge it at the same time as the ui, because the changes in `parent_id()` will break navigation (the _file_xxx ids will be mistaken for hashes and probably 404)

snoop/data/views.py
```diff
@@ -85,7 +97,10 @@ def trim_text(data):
 @collection_view
 def document(request, hash):
     digest = get_object_or_404(models.Digest.objects, blob__pk=hash)
-    return JsonResponse(trim_text(digests.get_document_data(digest)))
+    page = 1
+    if 'page' in request.GET:
+        page = request.GET['page']
```
gabriel-v, line 102: let's call these arguments `children_page`

gabriel-v, line 102: simpler: `page = int(request.GET.get('children_page', 1))`

snoop/data/digests.py
```diff
@@ -382,26 +392,31 @@ def child_dir_to_dict(directory):
...
-    child_directory_queryset = directory.child_directory_set.order_by('name_bytes')[:limit]
-    child_directory_queryset = child_directory_queryset[:limit]
+def get_directory_children(directory, page_index=1):
+    limit = settings.SNOOP_DOCUMENT_CHILD_QUERY_LIMIT
+    child_directory_queryset = directory.child_directory_set.order_by('name_bytes')
     child_file_queryset = directory.child_file_set.order_by('name_bytes')
-    child_file_queryset = child_file_queryset[:limit]
-    incomplete = len(child_directory_queryset) == limit or \
-        len(child_file_queryset) == limit
+    print(child_directory_queryset)
```
gabriel-v, line 404: please remove these print outs

generally i recommend adding `import pdb; pdb.set_trace()` (maybe through a single hotkey in your editor) and exploring the objects, instead of printing the <QuerySet ...> stuff around the code

snoop/data/digests.py
```diff
@@ -382,26 +392,31 @@ def child_dir_to_dict(directory):
...
-        [child_dir_to_dict(d) for d in child_directory_queryset][:limit - 1]
-        + [child_file_to_dict(f) for f in child_file_queryset][:limit - 1]
+        [child_dir_to_dict(d) for d in dir_page.object_list]
+        + [child_file_to_dict(f) for f in file_page.object_list]
     ), incomplete
 
 
-def get_directory_data(directory):
-    children, incomplete_children_list = get_directory_children(directory)
+def get_directory_data(directory, page=1):
```
gabriel-v, line 418: same as above, call the argument `children_page`

snoop/data/digests.py
```diff
@@ -414,3 +429,26 @@ def get_directory_data(directory):
         'children': children,
         'incomplete_children_list': incomplete_children_list,
     }
+
+
+def get_file_data(file, page=1):
```
gabriel-v, line 434: same as above, call the argument `children_page`

gabriel-v (changes requested):

gabriel-v:
merged in 0.12.0

PR #360, opened 2020-10-30
Non root user v1

gabriel-v:
can't comment if there's no PR


.drone.yml
```diff
@@ -28,41 +28,105 @@ steps:
...
+  volumes:
+  - name: snoop-pg-v
+    path: /drone/src/volumes/snoop-pg/data
+
 
 volumes:
 - name: dockersock
+  host:
+    path: /var/run/docker.sock
+- name: snoop-pg-v
```
gabriel-v, line 112: You're mounting all these things on the host, but we don't want to do that - that would mean all the test instances are running on the same data, which means they're not isolated tests.

Please remove all the stuff you added here, since the services only run once we don't really need to mount their volumes anywhere.

.drone.yml
```diff
@@ -28,41 +28,105 @@ steps:
 
 - name: download testdata
   image: alpine/git:1.0.7
+  volumes:
+  - name: test-collections-v
+    path: /drone/src/testsuite/volumes/collections
   commands:
   - mkdir -p testsuite/volumes/collections
   - git clone https://github.com/hoover/testdata testsuite/volumes/collections/testdata
+  - chown 666:666 testsuite/volumes/collections/testdata
```
gabriel-v, line 37: recursive

.drone.yml
```diff
@@ -28,41 +28,105 @@ steps:
...
 
-- name: py.test
-  image: docker/compose:1.24.0
+- name: run tests
+  image: liquidinvestigations/hoover-snoop2:test
+  pull: if-not-exists
   volumes:
   - name: dockersock
-    path: /var/run
+    path: /var/run/docker.sock
```
gabriel-v, line 52: don't need any volumes (maybe except for collections)

.drone.yml
```diff
@@ -28,41 +28,105 @@ steps:
...
+    POSTGRES_PASSWORD: snoop
+  volumes:
+  - name: snoop-pg-v
+    path: /drone/src/volumes/snoop-pg/data
+
 
 volumes:
 - name: dockersock
+  host:
+    path: /var/run/docker.sock
```
gabriel-v, line 111: You're making all tests use the same :test image on the host. What if we're running multiple tests on multiple branches at the same time?

Please leave this section as it was, maybe keep the `collections` volume

docker-entrypoint.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash -ex
+
+echo "Running in $PWD"
+
+mkdir -p /opt/hoover/snoop/testsuite/volumes/snoop-pg/data
```
gabriel-v, line 5: We don't want to know / control / think about any database data directory in this service.

docker-entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+chown $UID:$GID -R /opt/hoover/snoop/testsuite/volumes/snoop-pg/data
+
+chown -R 666:666 /runserver
```
gabriel-v, line 8: Change the owner of the constant files in the Dockerfile, not here

gabriel-v (commented):
please merge all of your 140 commits in this repo

PR #364, opened 2020-11-10
non-root-user

morten-spiegel:


.drone.yml
```diff
@@ -60,51 +85,98 @@ services:
...
-depends_on:
-- test
+- name: snoop-pg
+  image: postgres:12.2
+  environment:
+    POSTGRES_USER: snoop
+    POSTGRES_DATABASE: snoop
+    POSTGRES_PASSWORD: snoop
+  # volumes:
+  # - name: snoop-pg-v
```
gabriel-v, line 109: You can keep this temp volume here, since the service runs only once. I think it's better to have the data in a temp volume than on the final layer of the docker image, since that is overwritten in /opt as an overlayfs and the tmpfs will be a ramfs (much faster, no overhead).

.drone.yml
```diff
@@ -60,51 +85,98 @@ services:
...
-trigger:
-  status:
-  - success
-  event:
-  - push
-  - tag
+- name: search-es
+  image: docker.elastic.co/elasticsearch/elasticsearch:6.2.4
+  # volumes:
+  # - name: search-es-v
```
gabriel-v, line 97: same as below: you can keep the temp volume here

gabriel-v (changes requested):
Move the two publish steps from the final pipeline into the beginning of the first pipeline, so you can test the images you just built directly from docker hub.

So: remove any `image: docker` steps and `/var/run` volumes, since we don't want to do docker-in-docker anymore.

Here's the new step order:

- flake8
- build base docker image
- docker push branch ${DRONE_COMMIT_BRANCH}
- docker push autotag + latest
- download testdata
- run tests (image is liquidinvestigations/hoover-snoop2:${DRONE_COMMIT_BRANCH}) for the push event)

And remove the `docker daemon in docker` service.

.drone.yml
```diff
@@ -106,6 +63,89 @@ steps:
...
+steps:
+- name: download testdata
+  image: alpine/git:1.0.7
+  volumes:
+  - name: test-collections-v
+    path: /drone/src/testsuite/volumes/collections/testdata
+  commands:
+  - mkdir -p testsuite/volumes/collections
+  - git clone https://github.com/hoover/testdata testsuite/volumes/collections/testdata
+  - chown -R 666:666 testsuite/volumes/collections/testdata
```
gabriel-v, line 82: we shouldn't need this - default permissions for this volume should be that everyone can read the files. This is the case in production, too: the initial collection data will remain owned by `root:root`

gabriel-v (changes requested):
please rebase to master, the diff i'm looking at has unrelated changes

Also, please bring back temp volumes for all the volumes you have commented out in the drone job file.

PR #365, opened 2020-11-26
change to non root user in docker entry point

morten-spiegel:


.drone.yml
```diff
@@ -106,6 +63,89 @@ steps:
...
+  image: rabbitmq:3.7.3
+
+- name: snoop-tika
+  image: logicalspark/docker-tikaserver
+
+- name: search-es
+  image: docker.elastic.co/elasticsearch/elasticsearch:6.2.4
+  # volumes:
+  # - name: search-es-v
+  #   path: /drone/src/volumes/search-es/data
```
morten-spiegel, line 127: I am not sure which path I have to specify. All the examples for postgres and elasticsearch I can find never have volumes.

gabriel-v: That's because they fill up the root volume that comes with the container. Leave it like this for now, will revisit in the future if it becomes a problem...

docker-entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -ex
+
+if [[ ! -d "$DATA_DIR" ]]; then
+        exit 1
+fi
+
+chown $UID:$GID /opt/hoover
```
gabriel-v, line 7: Isn't this already set in the Dockerfile? Is this used as a direct mount point anywhere?

Dockerfile
```diff
@@ -12,10 +20,21 @@ COPY . .
...
  && echo '#!/bin/bash -e' > /runserver \
- && echo 'waitress-serve --threads $THREAD_COUNT --port 80 snoop.wsgi:application' >> /runserver \
+ && echo 'waitress-serve --threads $THREAD_COUNT --port 8080 snoop.wsgi:application' >> /runserver \
  && chmod +x /runserver
 
+RUN chown -R $UID:$GID /runserver
+RUN chown -R $UID:$GID /opt/magic-definitions
+RUN chown -R $UID:$GID /opt/libpst
+
+ENV DATA_DIR "/opt/hoover/snoop"
```
gabriel-v, line 30: This is only a mount point when mounted from liquid with `mount_local_repos`, a use-case which you shouldn't care about here.

The mount point we care about here is `/opt/hoover/snoop/blobs` which is filled with directories, one for each collection, with blobs. Call it "$BLOBS_DIR"

docker-entrypoint.sh
```diff
@@ -0,0 +1,11 @@
...
+chown -R $UID:$GID $DATA_DIR
```
gabriel-v, line 8: Please remove the recursive `chown` from the entrypoint: for large data dirs, this will take a very long time, even if there's no permission to change!

gabriel-v (changes requested):
I'd like to merge this in 2 parts:
- first part (this), that:
  - removes the docker-compose from testing
  - runs the container as 666:666 during testing, but allows `root:root` to run it in production too with no changes
- second part, that:
  - uses minio as a backend for blobs (minio will keep running as root:root)
    - some other issue & pull request somewhere around here, assign yourself :D

So, left in this part:
- remove comments and unused files (docker-compose stuff?)
- remove any `chown`s you don't need from the entrypoint, set all possible permissions from the dockerfile

docker-entrypoint.sh
```diff
@@ -0,0 +1,18 @@
...
+
+    echo "running as liquid"
+    exec gosu $USER_NAME "$@"
+
+else
+
+    echo "running as root"
+    exec "$@"
+
+fi;
```
morten-spiegel, line 18: About the two different merge parts. My Idea was to give a flag as an Env variable in Drone that we now run tests and if we get the flag we run as 666 if not we run as root. Its not working like this but do is this what you wanted or did I misunderstood that?

gabriel-v: Not really; if the container gets run as root by nomad then permissions don't matter to it anyway, right? So you shouldn't care what happens outside of the tests right now, and you may need to add an explicit `user = root` in the liquid job for these containers

gabriel-v:
merged into master / 0.13, thanks!

PR #410, opened 2021-02-05
[wip] add entity extraction model and serializer

gabriel-v:
closes https://github.com/CRJI/EIC/issues/297


snoop/data/digests.py
```diff
@@ -117,17 +118,24 @@ def gather(blob, **depends_on):
...
 
-    # combine texts and detect language
-    if settings.DETECT_LANGUAGE:
+    #TODO: Add entity Extraction setting to snoop settings
+    if settings.EXTRACT_ENTITIES:
+        text = rv.get('text', '')
+        ocrtexts = [v for v in rv.get('ocrtext', {}).values() if v]
+        if text or ocrtexts:
+            alltext = text + "\n" + "\n".join(ocrtexts)
+            nlp_response = entities.get_entities(alltext)
```
gabriel-v, line 127: Get the entities separately for each of the `text` and `ocrtext.*` because they are shown separately in the UI tab.


All your responses will also appear in the index (here, the `rv`) in fields keyed under the text identifier.

All the data you get from your service should be saved here into the Entity table before going forward.

snoop/data/digests.py
```diff
@@ -117,17 +118,24 @@ def gather(blob, **depends_on):
...
         rv['ocrtext'] = ocr_results
 
-    # combine texts and detect language
-    if settings.DETECT_LANGUAGE:
+    #TODO: Add entity Extraction setting to snoop settings
+    if settings.EXTRACT_ENTITIES:
+        text = rv.get('text', '')
+        ocrtexts = [v for v in rv.get('ocrtext', {}).values() if v]
+        if text or ocrtexts:
+            alltext = text + "\n" + "\n".join(ocrtexts)
```
gabriel-v, line 126: Don't concatenate these texts.

For language detection concatenate first 3K of each one like below, and for NER do it separately per text, as explained below.

.gitignore
```diff
@@ -1,4 +1,5 @@
 *.pyc
+*.code-workspace
```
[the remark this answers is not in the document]
MattesR: I can also use only one Workspace. I thought that VSCode only let's you have one venv per workpace, but that's actually not the case.

snoop/data/models.py
```diff
@@ -425,6 +425,13 @@ def delete(self, *args, **kwargs):
         digests.retry_index(self.blob)
 
 
+class Entity(models.Model):
+    entity = models.CharField(max_length=256, unique=True)
```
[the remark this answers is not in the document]
MattesR: 1. Yes, that's probably better than having a list of EntityIDs in each blob.
2. That's a good point, you're right.
3. Yeah it's the type. So another Table called EntityTypes and add it to migration.
4.Do we want the location in the Entity class?! Better to have it in EntityHit, no?

snoop/data/digests.py
MattesR: Okay. So:

- [ ] 1.  One Request per text (text and OCR text)
- [ ] 2.  Save all Entities in Entity table (unique Combination of label and text)
- [ ] 3. Store the start and end in EntityHit Table
One question about storing the stuff for EntityHit. What should I put in rv for it? Just all entities with all info?

snoop/data/models.py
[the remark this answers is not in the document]
gabriel-v: Yeah, put the location in the entity hit table, of course

Also respect database normalization everywhere else it pops up and create the tables you need

snoop/data/digests.py
gabriel-v: The keys `rv` go down to be served through the API and be stored to elasticsearch. This doesn't happen directly, there are some more functions that pick given fields to serve in the API or store in the index.

Here we always want to have the same data and structure, so let's think of some stuff we want in the index first for every document:

- `rv['entity']`: list of all the entities (de-duplicated) of all types (the interesting types, that we want in a single bucket - not dates)
- `rv['extracted-date']`: list of dates extracted from the document
- `rv['extracted-geolocation']`: list of GPS positions data extracted from the document (https://www.elastic.co/guide/en/elasticsearch/reference/6.8/geo-point.html to be used later with https://www.elastic.co/guide/en/elasticsearch/reference/6.8/geo-queries.html) - we need a local database of location names vs. GPS, so this may be out of scope for your project
- `rv['ent-type.X']`: for each entity type, a de-duplicated list of the texts
- `rv['ent-id.X']`: for each entity hit, put the numerical ID here


And in snoop also add these fields to the [mapping](https://github.com/liquidinvestigations/hoover-snoop2/blob/8a6dbdf1fd1ea56db386628f0559b097cdaa1a61/snoop/data/indexing.py#L22):
- `entity`: text (we'll use the `.keyword` generated sub-field for buckets, and use this for fuzzy searching done by the user, like `entity:Jim~2`)
- `extracted-date`: date
- `extracted-geolocation`: geo-point
- `ent-type`: object
- `ent-type.*`: text (see tags field for dynamic mapping, but use text instead of keyword)
- `ent-id.*`: int (same as above)

Be sure to forward these fields in [the ugly transcription function here](https://github.com/liquidinvestigations/hoover-snoop2/blob/8a6dbdf1fd1ea56db386628f0559b097cdaa1a61/snoop/data/digests.py#L489) that converts your `rv` (here called `digest_data`) and the state of the File objects into the stuff that gets indexed by es and presented on the API.

snoop/data/digests.py
```diff
@@ -117,17 +118,24 @@ def gather(blob, **depends_on):
                 rv['ocrimage'] = True
         rv['ocrtext'] = ocr_results
 
-    # combine texts and detect language
-    if settings.DETECT_LANGUAGE:
+    #TODO: Add entity Extraction setting to snoop settings
```
gabriel-v, line 121: to keep this function small, please move all code to a function inside your module:

```python
    rv.update(entities.extract(all, the, args, you, need, here))
```

and then in your module:

```python3


def extract(bla, bla):
    if settings.DETECT_LANGUAGE:
        ...
```

snoop/data/models.py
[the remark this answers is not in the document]
MattesR: > statically pre-populate in a django migration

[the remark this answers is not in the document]
MattesR: > I think we want another table EntityHit with unique ( Entity, Digest, text_id ) where text_id is text or ocrtext.blablabla

Where should the text ID come from? ForeignKey to blob as in `OcrDocument?`
I don't understand fully where text and OCR texts are stored per file. Is it in the blob(s)?! 
So an example: I index a pdf document `bla.pdf` in the root folder of my collection. It contains 3 OCR jobs/requests/parts and an extracted text. 
What is stored where in the db?

- 1 File, Presumably
- 4 blobs, and therefore 4 Digests?
- 1 OCRSource?  
- 3 OCRDocuments with the same source 
If so, I'd go ahead and add a ForeignKey to a blob to an entity hit.
If only one blob is created per file, where is the distinction between different OCR parts made?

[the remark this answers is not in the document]
gabriel-v: No, you get the OCR texts like the old language detection is getting them: keys called `text` and `ocrtext.*` on the `rv` object. 

You put your results back on the same `rv` object, and additionally create your own database structure for your own tables.

No need to worry about where it comes from or where it goes.

snoop/data/digests.py
```diff
@@ -142,15 +130,40 @@ def gather(blob, **depends_on):
...
     with models.Blob.create() as writer:
         writer.write(json.dumps(rv).encode('utf-8'))
 
-    _, _ = models.Digest.objects.update_or_create(
-        blob=blob,
-        defaults=dict(
-            result=writer.blob,
-        ),
-    )
+        digest.result = writer.blob
```
gabriel-v, line 165: this needs to be outside the `as writer` block, see the original code: 

```python
with models.Blob.create() as writer:
    writer.write(json.dumps(rv).encode('utf-8'))

# rest of the stuff on this same level here, not in block above:
...
```

The "writer" is a smart context manager that only creates the `.blob` on exiting the context.

snoop/data/digests.py
```diff
@@ -142,15 +130,40 @@ def gather(blob, **depends_on):
...
+                    else:
+                        rv.update(ocr_results)
+    elif settings.DETECT_LANGUAGE:
+        text = rv.get('text', '')[:2500]
+        if text:
+            rv.update(entities.get_language(text))
+        if ocr_results:
+            for ocr_name, ocrtext in rv.get('ocrtext'):
+                if ocrtext:
+                    rv[f'{ocr_name}_lan'] = entities.get_language(ocrtext[:2500])
```
gabriel-v, line 160: Let's move all this noise to a single function `process_document` inside your module, so here you'd just call:
`rv.update(entities.process_document(digest, rv))`

I want to keep this function `gather` as simple as possible.

snoop/data/digests.py
```diff
@@ -142,15 +130,40 @@ def gather(blob, **depends_on):
             rv['location'] = exif_data.get('location')
             rv['date-created'] = exif_data.get('date-created')
 
+    digest, _ = models.Digest.objects.get_or_create(blob=blob)
+    rv = defaultdict([], rv)
```
gabriel-v, line 134: why the copy here? who's gonna reference the old one, anyway?

snoop/data/analyzers/entities.py
```diff
@@ -0,0 +1,63 @@
...
+        results['lang'] = nlp_response['language']
+    results['entities'] = [k['text'] for k in ents]
+    unique_ents = set([(k['text'], k['type']) for k in ents])
+    results['ent-ids'] = [models.Entity.objects.get(entity=k[0], type=k[1]).id
+                          for k in unique_ents]
+    for entity_type in set(v['type'] for v in results['entities']):
+        results[f'entity-type.{entity_type}'] = [k[0] for k in unique_ents
+                                                 if k[1] == entity_type]
+
+    if settings.DETECT_LANGUAGE and 'lang' not in results:
```
gabriel-v, line 60: `lang` will always be `in results` because you set it yourself a couple of lines above. `'key' in X` will be True even if `X['key'] ` is None...


I think it's best to respect the comment you made in `gather()` and always do language detection if we're extracting entities. This means you remove the `if` above too. (That one's funny too: if we have the language code ID, why wouldn't we set it on the result?)

snoop/data/models.py
```diff
@@ -425,6 +426,42 @@ def delete(self, *args, **kwargs):
...
+    description = models.CharField(max_length=256, unique=True)
+
+    def __str__(self):
+        return self.description
+
+
+class EntityHit(models.Model):
+    entity = models.ForeignKey(Entity, on_delete=models.CASCADE)
+    digest = models.ForeignKey(Digest, on_delete=models.CASCADE)
+    model = models.ForeignKey(LanguageModel, on_delete=models.RESTRICT)
```
gabriel-v, line 459: set this to `CASCADE` - if the model dissapears, then we want all related to it to dissapear too without problems.

snoop/data/migrations/0041_auto_20210215_1046.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('data', '0040_auto_20210215_1045'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_entity_types),
```
gabriel-v, line 27: Set NoOP for reverse of migration (see other RunPython migrations around here), so we can reverse it.

snoop/data/models.py
```diff
@@ -425,6 +426,42 @@ def delete(self, *args, **kwargs):
         digests.retry_index(self.blob)
 
 
+class EntityType(models.Model):
+    type = models.CharField(max_length=256, unique=True)
```
gabriel-v, line 430: Are we sure we want to implement custom entity types created by the user in this way? Without some kind of training phase and/or predictive coding, this seems kind of useless and will only complicate the code. Sure, you can have a rule to re-tag previously set "NAZI" type entities with the same custom type, but it won't follow the context: if the same text for your NAZI entity is used in some other context (where it's the name of a place or a work of art or something else), then you'll have a wrong match.

Maybe what we want here is for people to be able to tag some existing (entity, type) pair with their own tags, so they'd say `PERSON: Adolf Hitler` has the "NAZI" tag - and we'd tag the whole document with `entity-custom-tags: ['NAZI']`


If we do it that way, then we don't want a whole table for this when it's only used as a Foreign Key on the Entity. It's going to be annoying to manage this from python code; you have to add more data migrations to append values to this list.

Let's use a multiple choice string field on the `Entity.type` (same as in `snoop.data.models.Tasks.status`) and remove this table and its data migration, then add another table called "CustomEntityType" or "CustomEntityTag" or something, with a Many to Many relationship with "Entity"

gabriel-v (commented):
so far so good. 

First, rebase over `master` and add docstrings to all your new code. 

Then, there's one big thing you need to add: for all the models you make, create Admin site views for them in `snoop.data.admin` and register them on the admin site. You should make most / all fields read only; explore the rest of the admin site to get an idea.

snoop/data/analyzers/entities.py
```diff
@@ -0,0 +1,63 @@
...
+
+def get_language(text):
+    data = {'text': text}
+    return call_nlp_server('language_detection', data)['language']
+
+
+# Create all db entries which are related to that entity.
+def create_db_entries(entity, model, language, text_source, digest):
+    db_ent = models.Entity.objects.get_or_create(entity=entity['text'], type=entity['type'])
+    models.EntityHit.objects.get_or_create(entity=db_ent,
```
gabriel-v, line 33: use the coding style here which doesn't waste half of the screen pls

```python
    models.EntityHit.objects.get_or_create(
        entity=db_ent,
```

snoop/data/models.py
MattesR: Custom entity types are pretty important, but I don't think that we'll need something as complicated as predicting these custom entity types. It's enough to just match the ones ppl add to the list of entities of their custom type entity list documents. Since ppl need to be able to add entities to existing types as well, there's not much of a difference.
Let's say you have a drug trafficking ring or something who operates with animal names as code names. First, it would be helpful to have a custom entity type `RING` or something, and maybe you want to consider `owl` as an entity in this context. 
We should be able to look for these custom entities just like for the others.

snoop/data/digests.py
MattesR: I've casted `rv` into a `defaultdict` in order to append to the entities if entities are found in the text as well as in the OCR results. But since I've put all that into the entities file, I only need the `defaultdict` in there now.

snoop/data/models.py
MattesR: > going to be annoying to manage this from python code; 
> you have to add more data migrations to append values to this list.

Wait, I don't really get that. Why do we need new migrations if we add to the list? Can't we just use `EntityType.objects.get_or_create(type='NAZI')` as we do with all other parts of the model?

snoop/defaultsettings.py
```diff
@@ -169,12 +169,8 @@
...
 
 DETECT_LANGUAGE = True
-"""Enable language detection."""
-
-LANGUAGE_DETECTOR_NAME = 'polyglot'
-"""Configure which language detector library to use.
-"""
-
+EXTRACT_ENTITIES = True
+NLP_SERVICE_URL = os.environ.get('SNOOP_NLP_URL', 'http://127.0.0.1:5000/')
```
gabriel-v, line 173: docstrings for all 3 of these pls.

remember docstrings for assignments are after the assignment (see rest of this file)

snoop/data/models.py
gabriel-v: Ok, then let's do `EntityType.objects.get_or_create` everywhere and remove the migration that adds the rows with classes. The classes will be added when they appear in the data, and we won't have any "magic" classes that come pre-created in the database.

PR #425, opened 2021-03-05
Added first version of filetypes table

k-jell:
closes CRJI/EIC#667

This is the first version of the table. Maybe the supported file extensions should also be added as a column?


filetypes.md
```diff
@@ -0,0 +1,56 @@
...
+| application/zip                                                           | archive         | archives       |
+| application/rar                                                           | archive         | archives       |
+| application/x-rar                                                         | archive         | archives       |
+| application/x-zip                                                         | archive         | archives       |
+| application/x-gzip                                                        | archive         | archives       |
+| application/gzip                                                          | archive         | archives       |
+| application/x-bzip2                                                       | archive         | archives       |
+| application/x-tar                                                         | archive         | archives       |
+| application/x-hoover-pst                                                  | email-archive   | archives       |
+| application/mbox                                                          | -               | archives       |
```
gabriel-v, line 13: let's map this to `email-archive` in the code and here too. Does mapping it in the code change any logic? if it does, we want to refactor it later so the logic doesn't depend on these values...

filetypes.md
```diff
@@ -0,0 +1,56 @@
...
+| application/vnd.oasis.opendocument.presentation-template                  | ppt             | tika           |
+| application/xhtml+xml                                                     | -               | html           |
+| application/xml                                                           | -               | html           |
+| text/xml                                                                  | -               | htlm           |
+| image/tiff                                                                | image           | exif, ocr      |
+| image/jpg                                                                 | image           | exif, ocr      |
+| image/webp                                                                | image           | exif, ocr      |
+| image/heic                                                                | image           | exif, ocr      |
+| application/vnd.ms-outlook                                                | email           | email          |
+| application/vnd.ms-office                                                 | -               | email          |
```
gabriel-v, line 53: i suppose `application/vnd.ms-office` and `application/CDFV2` are micro$ email formats -let's map them to `email` too

filetypes.md
```diff
@@ -0,0 +1,56 @@
+| Mime type                                                                 | Hoover Filetype | Analyzers      |
+| ---                                                                       | ---             | ---            |
+| application/x-7z-compressed                                               | archive         | archives       |
```
gabriel-v, line 3: Please add another column called `About` where you display a pretty name / explanation for the mime type. You can pull these from the mozilla list or wikipedia or anywhere else online


- https://www.freeformatter.com/mime-types-list.html
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

gabriel-v (changes requested):
thanks for the research, let's try and make some code changes here too since we found holes in our filetype mapping...

PR #427, opened 2021-03-09
Management command to display filetype statistics for a collection

k-jell:
closes CRJI/EIC#667

This should display the correct information.


snoop/data/management/commands/filestats.py
```diff
@@ -0,0 +1,128 @@
...
+from django.core.management.base import BaseCommand
+from ...logs import logging_for_management_command
+from ... import models
+from ... import collections
+from django.db.models import Count
+from django.db.models import Sum
+import re
+import mimetypes
+
+SUPPORTED_FILETYPES = {
```
gabriel-v, line 11: Actually, I think you can re-create this whole list from the code base itself. Just import all of the respective modules and concatenate the lists inside. For places where the content type isn't extracted as a global constant string or list, then refactor it and add a reference to it in this file.

This will let us maintain this command in the long term without actually visiting it.

snoop/data/management/commands/filestats.py
```diff
@@ -0,0 +1,128 @@
...
+    'image/heic',
+
+    'application/vnd.ms-outlook',
+    'application/vnd.ms-office',
+    'application/CDFV2',
+    'message/rfc822',
+    'message/x-emlx',
+}
+
+def truncateSize(size):
```
gabriel-v, line 76: use `round(size, -(len(str(size)) - 1))` or something https://docs.python.org/3/library/functions.html#round

gabriel-v, line 76: coding style in this code base is `under_scores` not `javaScript` thank you

https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

https://www.python.org/dev/peps/pep-0008/#prescriptive-naming-conventions

snoop/data/management/commands/filestats.py
```diff
@@ -0,0 +1,128 @@
...
+            for mtype in querysetMime:
+                print(f'{(mtype["mime_type"]):75} {truncateSize(mtype["size"]):10d}')
+
+            print()
+            print('---- Top Unsupported Mime-Types ----')
+            print()
+            unsuppQuery = querysetMime.exclude(mime_type__in=SUPPORTED_FILETYPES)
+            for mtype in unsuppQuery[:100]:
+                #if mtype['mime_type'] not in SUPPORTED_FILETYPES:
+                extensions = mimetypes.guess_all_extensions(mtype['mime_type'])
```
gabriel-v, line 105: here I want to get the sizes for the actual extensions from the actual files on disk from the `blob.file_set`

the filename field you have to split is binary, see here: https://github.com/liquidinvestigations/hoover-snoop2/blob/10941701ea6bda5a8efe27bcceb48cce19616a4f/snoop/data/models.py#L186

you probably want to use these nasty functions here https://docs.djangoproject.com/en/3.1/ref/models/database-functions/#django.db.models.functions.StrIndex that translate to their nasty SQL counterparts for doing this; i'm sure they work just fine on binary fields too (just be sure to give them `b'...'` strings)

please keep this if possible to one single query (like the ones for mime-type) since looping in python between 2 different queries will take forever on a 2TB dataset. also please display the extensions one per line, sorted after size, while also displaying the mime-type for each extension

gabriel-v (changes requested):

snoop/data/management/commands/filestats.py
gabriel-v: or just do this

```
size_for_ext = defaultdict(int)
mime_for_ext = defaultdict(set)
for f in File.objects.all().iterator():
    ext = f.name_bytes.split(b'.')[-1][:100]
    size_for_ext[ext] += f.size
    mime_for_ext[ext].add(f.mime_type)

print(size_for_ext)
print(mime_for_ext)

File: snoop/data/management/commands/filestats.py [hunk 4]
```diff
@@ -0,0 +1,128 @@
...
+    def handle(self, collection, **options):
+
+        col = collections.ALL[collection]
+        
+        with col.set_current():
+            querysetMime = models.Blob.objects.all().values('mime_type').annotate(total=Count('mime_type')).annotate(size=Sum('size')).order_by('-size')
+            print()
+            print('---- Top 100 Mime-Types by size ----')
+            print()
+            for mtype in querysetMime:
```
gabriel-v, line 96: make some top-level functions in this file called `get_top_mime_types()` and `get_top_extensions()` that return python dicts with the data you want to print() in this command here. This has the added benefit we can later just move your functions somewhere else in the code base and display this info in the front-end too.

snoop/data/management/commands/filestats.py
```diff
@@ -0,0 +1,128 @@
...
+def truncateSize(size):
+    exponent = len(str(size))-1
+    divisor = 10**exponent
+    return int((size/divisor))*divisor
+
+class Command(BaseCommand):
+    help = "Display filetype stats."
+
+    def add_arguments(self, parser):
+        parser.add_argument('collection', type=str)
```
gabriel-v, line 85: instead of running this for only one collection, let's combine all collections on one server. I wouldn't wanna recombine these on my own from copy/pasted texts, right?

if you want you can set up an optional argument that you can specify multiple times called `collection` - when you have that argument, you only run it on the listed collections, not on all of them. `./manage.py filestats --collection testdata --collection testdata2`

it's fine if you collect top 100 results for each collection, then recombine the results (using dicts and sets like below) and then show a combined list that's actually longer than 100

gabriel-v (changes requested):
some small formatting changes to go.

Let's run this management command at the end of the integration test, so it can fail if anything is wrong.

Put a couple more lines at the bottom of this test:

https://github.com/liquidinvestigations/hoover-snoop2/blob/08695acc98d71d06f557d0c912585d4c827ce993/testsuite/test_integration.py#L112

At top of file:
```python
from snoop.data.management.commands import get_top_mime_types, get_top_extensions
```
Then, at bottom of test:

```python
mime_dict_supported = get_top_mime_types(['testdata'], True)
assert 'application/pdf' in mime_dict_supported.keys()
mime_dict_unsupported = get_top_mime_types(['testdata'], False)
assert 'application/pdf' not in mime_dict_unsupported.keys()

ext_dict1 = get_top_extensions(['testdata'], True)
# some simple asserts here too
ext_dict2 = get_top_extensions(['testdata'], False)
```

snoop/data/management/commands/filestats.py
```diff
@@ -0,0 +1,181 @@
...
+    return dict(sorted_ext_dict)
+
+
+def get_description(col, mime_type, *extension):
+    """Return the magic description for a given mime-type.
+
+    Args:
+        col: Collection on which the query is executed.
+        mime_type: Mime-Type for which the descriptions is returned.
+        *extension: Optional file-extension as string to limit the search to this
```
gabriel-v, line 103: 1. I don't see anywhere where this `*extension` argument is used -- you didn't add the mime descriptions for the second extension table, and that's OK. Since the argument is unused, I'd delete the argument and the `if extension` and simplify this function.
2. `*extension` doesn't mean optional argument, it means variadic argument list. You generally don't want to use variadic functions unless you've got some good reason to (like composition when being called by some library with arguments you don't care about, or when it's natural inside the caller code) -- here, there's no reason to: you seem to want to get an optional list of extensions (only to ignore most of them and look at the first one in the list) -- so you should use an argument with default value  `extensions=[]`. But then, if you somehow modify the list, you run into this: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments - so either do the work-around there (leave the default argument value as `=None` and then change it), or be sure not to mutate the argument. Anyway, you probably don't need this argument anyway, but I wanted you to know the problems here

snoop/data/management/commands/filestats.py
```diff
@@ -0,0 +1,181 @@
...
+        collection_args = list(collections.ALL.keys())
+        supported = True
+        if options['unsupported']:
+            supported = False
+        if options['collections']:
+            collection_args = options['collections']
+
+        print('Top Mime Types by size')
+        print('-----------------------')
+        for k, v in get_top_mime_types(collections_list=collection_args, print_supported=supported).items():
```
gabriel-v, line 169: here do `size =  v["size"] / ( 2 ** 20 ))` and display this float with 2 digits after the `.` and the `MB` unit after

i'm getting dizzy from all the long numbers, so please format the float to also include the `,` every 3 digits so we easily see when we pass into the GB range

snoop/data/management/commands/filestats.py
```diff
@@ -0,0 +1,181 @@
...
+    ext_dict = {}
+    for col in collections_list:
+        query = r"""select substring(encode(f.name_bytes::bytea, 'escape')::text
+                    from '(\..{1,20})$') as ext,
+                    sum(f.size) as size,
+                    b.mime_type as mime
+                    from data_file f
+                    join data_blob b on f.blob_id = b.sha3_256
+                    group by ext, mime
+                    order by size desc limit 100;"""
```
gabriel-v, line 79: Put the 100 in a new argument `--row-count`, i want to run it on 1000 extensions too.

snoop/data/management/commands/filestats.py
```diff
@@ -0,0 +1,181 @@
...
+            action='store_true',
+            help='exclude supported filetypes')
+
+        parser.add_argument(
+            '--descriptions',
+            action='store_true',
+            help='print MIME-type descriptions')
+
+        parser.add_argument(
+            '--full_descriptions',
```
gabriel-v, line 145: rename this argumnent with `-` instead of `_` so you call it with `--full-descriptions`. You can add another argument to change its name in python to avoid doing `args["full-descriptons"]`, see docs or other management commands around here.

snoop/data/management/commands/filestats.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class Command(BaseCommand):
+    """Print the statistics for mimetypes or file-extendsion"""
+    help = "Display filetype stats."
+
+    def add_arguments(self, parser):
+        """Arguments to show only unsupported types, include magic descriptions,
+        include full magic descriptions and for choosing specific collections"""
+
```
gabriel-v, line 133: Add a new numeric argument called `--row-count` default 100 that you'll use to limit the number of rows in each request.

snoop/data/management/commands/filestats.py
```diff
@@ -0,0 +1,181 @@
...
+    res = {}
+    for col in collections_list:
+        collection = collections.ALL[col]
+        with collection.set_current():
+            queryset_mime = models.Blob.objects.all().values('mime_type', 'magic')\
+                .annotate(total=Count('mime_type')).annotate(size=Sum('size'))\
+                .order_by('-size')
+            if not print_supported:
+                queryset_mime = queryset_mime.exclude(mime_type__in=SUPPORTED_MIME_TYPES)
+            for mtype in queryset_mime:
```
gabriel-v, line 53: Stop after the top `--row-count` / 100 mime types for each request:

```python3
queryset_mime = queryset_mime[:row_count]
```

You should also rearrange the lines above so you do the `.exclude` first, then the `values`, `annotate`, then `order_by`, and last is the limit `[:row_count]`. Django may complain because you're doing `.exclude` after `.order_by`; check docs.

snoop/data/management/commands/filestats.py
```diff
@@ -0,0 +1,181 @@
...
+            elif options['full_descriptions']:
+                print(f'{k:75} {v["size"]:12d} {str(v["magic"])}')
+            else:
+                print(f'{k:75} {v["size"]:12d}')
+
+        print()
+        print('Top File Types by size')
+        print('-----------------------')
+        for k, v in get_top_extensions(collections_list=collection_args, print_supported=supported).items():
+            print(f'{str(k):75} {v["size"]:12d} {v["mtype"]}')
```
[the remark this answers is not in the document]
gabriel-v: Also, for the total size: same as above, display the total in MB with `.2f` and `,` every 3 digits.

gabriel-v (changes requested):
Let's make the row count configurable too; you can't really explore datasets if you get truncated at some count you can't control.

Everything else looks really good - only some small formatting changes to go

gabriel-v (approved):

PR #435, opened 2021-03-18
Map missing filetypes

k-jell:
closes CRJI/EIC#667.

Related to #425.

I mapped the missing filetypes. For `application/vnd.ms-office` and `application/CDFV2` I couldn't really find any information on what they actually are but they are mapped to `email` for now.

The only logic that is affected by the mapped type seems to be here:
https://github.com/liquidinvestigations/hoover-snoop2/blob/99b55e5c69a1bd522136effedb886f8d2bf667b8/snoop/data/digests.py#L445-L458

Do we want any changes there?

### DIFF
--- a/snoop/data/_file_types.py
+++ b/snoop/data/_file_types.py
@@ -5,13 +5,9 @@ FILE_TYPES = {
     'text/html': 'html',
     'application/x-hush-pgp-encrypted-html-body': 'html',
     'application/xhtml+xml': 'html',
-    'application/xml': 'html',
-    'text/xml': 'html',
     'message/x-emlx': 'email',
     'message/rfc822': 'email',
     'application/vnd.ms-outlook': 'email',
-    'application/vnd.ms-office': 'email',
-    'application/CDFV2': 'email',
 
     'application/x-hoover-pst': 'email-archive',
     'application/mbox': 'email-archive',
--- a/snoop/data/_file_types.py
+++ b/snoop/data/_file_types.py
@@ -4,11 +4,17 @@ FILE_TYPES = {
     'text/plain': 'text',
     'text/html': 'html',
     'application/x-hush-pgp-encrypted-html-body': 'html',
+    'application/xhtml+xml': 'html',
+    'application/xml': 'html',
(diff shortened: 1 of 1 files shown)


gabriel-v:
I found some notes explaining what's up:
- `application/vnd.ms-office` and `application/CDFV2` are microsoft pre-DOCX era (.doc) containers. They can be anything - outlook (single email or collection), excel, powerpoint, or anything else from their Office suite. Let's leave them unmapped, since they can hit anything
- `application/xml` and `text/xml` will probably be databases or some other resources, not XML - let's leave these two unmapped

Let's merge the management command first, and then we'll see what kinds of extensions we get for those mime types.

snoop/data/_file_types.py
```diff
@@ -4,11 +4,17 @@
     'text/plain': 'text',
     'text/html': 'html',
     'application/x-hush-pgp-encrypted-html-body': 'html',
+    'application/xhtml+xml': 'html',
+    'application/xml': 'html',
+    'text/xml': 'html',
```
gabriel-v, line 9: leave unmapped, as stated in previous comment

snoop/data/_file_types.py
```diff
@@ -4,11 +4,17 @@
...
     'message/x-emlx': 'email',
     'message/rfc822': 'email',
     'application/vnd.ms-outlook': 'email',
+    'application/vnd.ms-office': 'email',
```
gabriel-v, line 13: also leave these unmapped

gabriel-v (changes requested):
as i said in the previous comment, let's leave the only new mappings:

- `application/xhtml+xml`
- and `application/mbox`

All the others could be something else... we'll see what to do with them some other time.

I'll open some new issues about doing something about these specific formats, but that's not really a priority right now since they already work (when sent to Tika).


We can probably get more information out of Tika to begin with - this sounds really fun to do, so i'll assign this to you too @k-jell

snoop/data/_file_types.py
k-jell: Changed it.

PR #440, opened 2021-03-24
refactor archive_walk using os.scandir

k-jell:
closes CRJI/EIC#670

There is still an issue, where the temporary directories are included in the created path.

### DIFF
--- a/snoop/data/analyzers/archives.py
+++ b/snoop/data/analyzers/archives.py
@@ -9,9 +9,6 @@ import re
 from ..tasks import snoop_task, SnoopTaskBroken, returns_json_blob
 from .. import models
 import os
-import pprint
-from timeit import default_timer as timer
-import sys 
 
 SEVENZIP_MIME_TYPES = {
     'application/x-7z-compressed',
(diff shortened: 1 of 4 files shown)


snoop/data/analyzers/archives.py
```diff
@@ -195,21 +193,28 @@ def unarchive(blob):
...
-                    list(archive_walk(thing)),
-                    key=lambda c: c['name'],
-                ),
+    res = []
+    for root, dirs, files in os.walk(path, topdown=True):
+        children = []
+        for d in dirs:
+            dir_info = {
+                'type' : 'directory',
+                'name' :  os.path.relpath(d),
```
gabriel-v, line 202: this should be only the last path segment (in your case `d.name` - look at the code you deleted) and not some path:

```                
                'type': 'directory',
                'name': thing.name,
```

snoop/data/analyzers/archives.py
```diff
@@ -195,21 +193,28 @@ def unarchive(blob):
...
+        for f in files:
+            file_info = {
+                'type' : 'file',
+                'name' : os.path.relpath(f),
+                'blob_pk' : models.Blob.create_from_file(Path(os.path.join(root, f))).pk,
             }
+            children.append(file_info)
+        root_info = {
+            'type' : 'directory',
+            'name' : os.path.relpath(root),
```
gabriel-v, line 215: same as above: this should be the root name of object (for the root directory it's probably `''` the empty string)

snoop/data/analyzers/archives.py
```diff
@@ -195,21 +193,28 @@ def unarchive(blob):
...
-        else:
-            yield {
-                'type': 'file',
-                'name': thing.name,
-                'blob_pk': models.Blob.create_from_file(thing).pk,
+            children.append(dir_info)
+        for f in files:
+            file_info = {
+                'type' : 'file',
+                'name' : os.path.relpath(f),
```
gabriel-v, line 209: same as above: this should be the name of the file, not the path

gabriel-v (changes requested):

snoop/data/analyzers/archives.py
```diff
@@ -196,20 +194,28 @@ def unarchive(blob):
...
                 'type': 'file',
-                'name': thing.name,
-                'blob_pk': models.Blob.create_from_file(thing).pk,
+                'name': entry.name,
+                'path': entry.path
             }
+
+
+def create_blobs(dirlisting):
+    """Create blobs for files in archive listing created by archive_walk"""
```
gabriel-v, line 213: Add links like this: `[archive_walk][snoop.data.analyzers.archive_walk]`.

End docstring with a `.` because that's what the docstring style looks like.

gabriel-v (changes requested):
Looking good so far.


In `unarchive()` let's change the line `with tempfile.TemporaryDirectory() as temp_dir` to place the temporary directory under the Blobs temp dir and not in `/tmp`. The temp dir I'm talking about is this one: https://github.com/liquidinvestigations/hoover-snoop2/blob/a4b924d3dcbc316f3a391e64cd551afce2958f07/snoop/data/models.py#L178

I think it's best you made that `tmp` directory a property under the `Collection` class called `tmp_dir`. You can move the `blob_root()` function into the `collection.py` file too, as another property on `Collection` called `blobs_dir`, to avoid any circular imports.

Then, you can change all the places where these paths appear with `collections.current().tmp_dir` and `collections.current().blobs_dir`.

gabriel-v:
You have one failing test, see https://jenkins.liquiddemo.org/liquidinvestigations/hoover-snoop2/1816/3/7

```
FAILED testsuite/test_archives.py::test_unarchive_mbox - AssertionError: asse...
```

and if you scroll up

```



```

gabriel-v (approved):
LGTM, fix that test and we're merging this

gabriel-v (approved):

PR #442, opened 2021-04-16
Thumbnails

k-jell:
Closes CRJI/EIC#711


gabriel-v:
your tests are failing:

https://jenkins.liquiddemo.org/liquidinvestigations/hoover-snoop2/1835/3/7

```


ERROR    snoop.data.tasks:tasks.py:376 #4175 thumbnails.get_thumbnail(['c85700df5fc73c306b6367e5c6a66d69cb3bc275fc370eb011510d50ad47f95e']) [error] failed: TypeError("unsupported operand type(s) for +: 'NoneType' and 'str'") [0.000 s]
--



```


You have to:
- make everything work correctly even if the thumbnail service is not there
  - if you don't have the service, just don't crate the extra tasks!
- configure the env with thumbnail service when testing
  - there's another settings file in the testsuite


Fix the tests and then i'll review.

snoop/data/migrations/0041_thumbnail_size.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.1.4 on 2021-04-16 16:17
```
gabriel-v, line 1: Please collapse these 3 migrations into a single one:
- delete them both
- re-run makemigrations

snoop/data/migrations/0040_thumbnail.py
```diff
@@ -0,0 +1,22 @@
...
+        ('data', '0039_auto_20210204_2125'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Thumbnail',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('blob', models.OneToOneField(on_delete=django.db.models.deletion.CASCADE, to='data.blob')),
+                ('original', models.OneToOneField(on_delete=django.db.models.deletion.CASCADE, to='data.digest')),
```
gabriel-v, line 19: Why do we need the `original` here?

Just run the service on the `digest.blob` object and ignore the `original` stuff.

snoop/data/models.py
```diff
@@ -882,3 +882,27 @@ def __str__(self):
...
+class Thumbnail(models.Model):
+    """Database model for storing the Thumbnail corresponding to a Digest.
+    """
+
+    class Size(models.IntegerChoices):
+        SMALL = 100
+        MEDIUM = 200
+        LARGE = 400
+
+    original = models.ForeignKey(
```
gabriel-v, line 896: Again, why is this needed? you can pull the Digest from the `blob` with no issues.

Let's not complicate the schema any more than we need to.

You probably want to take a look at these: https://en.wikipedia.org/wiki/Database_normalization 2NF 3NF

snoop/data/urls.py
```diff
@@ -23,4 +23,5 @@
     path('<collection>/<hash>/raw/<filename>', views.document_download),
 
     path('<collection>/<hash>/tags/<username>/<uuid>', include(tags_router.urls)),
+    path('<collection>/thumbnail', views.thumbnail)
```
gabriel-v, line 26: make the URL look like the other ones:
- use the `<hash>` parameter in the URL instead of a `get` parameter called `digest`
- put the size in the URL too

Generally, it's really helpful to have the extension in the URL too (even if not required).

So make the URL look something like this instead:
`<collection>/<hash>/thumbnail/100.jpg`

snoop/data/views.py
```diff
@@ -312,3 +312,14 @@ def destroy(self, request, pk=None, **kwargs):
...
+
+
+@collection_view
+def thumbnail(request):
+    digest_sha3 = request.GET['digest']
+    size = request.GET['size']
+    digest_id = models.Digest.objects.get(blob_id=digest_sha3).pk
+    thumbnail_obj = models.Thumbnail.objects.get(original=digest_id, size=size)
+    thumbnail_blob = thumbnail_obj.blob
+    thumbnail_data = thumbnail_blob.open().read()
```
gabriel-v, line 324: use the `with blabla.open():` because python won't close the file for you

snoop/data/views.py
```diff
@@ -312,3 +312,14 @@ def destroy(self, request, pk=None, **kwargs):
         """Collection-aware overload that also checks permission to write tag."""
         self.check_ownership(pk)
         return super().destroy(request, pk, **kwargs)
+
+
+@collection_view
+def thumbnail(request):
+    digest_sha3 = request.GET['digest']
+    size = request.GET['size']
+    digest_id = models.Digest.objects.get(blob_id=digest_sha3).pk
```
gabriel-v, line 321: Use `get_or_404` instead of `get` in both these lines here.

gabriel-v (changes requested):
Your tests are failing, please fix them so they pass.

You need to make this work when the Thumbnail service is missing (configuration emtpy). In that case, just avoid starting the task from `digests.launch()` if the service is not set.

snoop/data/views.py
```diff
@@ -312,3 +312,14 @@ def destroy(self, request, pk=None, **kwargs):
...
+    thumbnail_obj = models.Thumbnail.objects.get(original=digest_id, size=size)
+    thumbnail_blob = thumbnail_obj.blob
+    thumbnail_data = thumbnail_blob.open().read()
+    return HttpResponse(thumbnail_data, content_type='image/jpeg')
```
gabriel-v, line 325: You should avoid storing the pic content in python RAM when an alternative is available, probably something like:
```
with blob.open() as f:
    return HTTPResonse(f, content_type='image/jpeg')
```

snoop/data/views.py
```diff
@@ -312,3 +312,14 @@ def destroy(self, request, pk=None, **kwargs):
...
         self.check_ownership(pk)
         return super().destroy(request, pk, **kwargs)
+
+
+@collection_view
+def thumbnail(request):
+    digest_sha3 = request.GET['digest']
+    size = request.GET['size']
+    digest_id = models.Digest.objects.get(blob_id=digest_sha3).pk
+    thumbnail_obj = models.Thumbnail.objects.get(original=digest_id, size=size)
```
gabriel-v, line 322: you can collapse the two DB queries into one: `models.Thumbnail.objects.get(size=size, original__pk=digest_sha3)`

snoop/data/digests.py
```diff
@@ -170,13 +175,30 @@ def gather(blob, **depends_on):
...
         ),
     )
 
+    # get thumbnail and create new thumbnail entry in the database
+    thumbnail_blobs = depends_on.get('get_thumbnail')
+
+    if thumbnail_blobs:
+        if isinstance(thumbnail_blobs, SnoopTaskBroken):
+            rv['broken'].append(thumbnail_blobs.reason)
+            log.debug("get_thumbnail task is broken; skipping")
```
gabriel-v, line 191: Also add a boolean flag here called `has-thumbnails` that is `False` unless otherwise set.

snoop/data/views.py
```diff
@@ -312,3 +312,10 @@ def destroy(self, request, pk=None, **kwargs):
         """Collection-aware overload that also checks permission to write tag."""
         self.check_ownership(pk)
         return super().destroy(request, pk, **kwargs)
+
+
+@collection_view
+def thumbnail(request, hash, size):
+    # digest = get_object_or_404(models.Digest.objects, blob__pk=hash)
```
gabriel-v: remove commented code

testsuite/test_thumbnail.py
```diff
@@ -0,0 +1,47 @@
...
+        blob = fakedata.blob(f.read())
+
+    fakedata.file(root, 'file.doc', blob)
+
+    taskmanager.run()
+
+    api = CollectionApiClient(client)
+    digest = api.get_digest(blob.pk)['content']
+
+    print(digest.keys())
```
gabriel-v, line 27: remove this print

testsuite/test_thumbnail.py
```diff
@@ -0,0 +1,47 @@
...
+    with test_pdf.open('rb') as f:
+        blob = fakedata.blob(f.read())
+
+    fakedata.file(root, 'file.pdf', blob)
+
+    taskmanager.run()
+
+    api = CollectionApiClient(client)
+
+    sizes = [100, 200, 400]
```
gabriel-v, line 44: take values from models, don't repeat yourself

testsuite/test_thumbnail.py
```diff
@@ -0,0 +1,47 @@
...
+from snoop.data.analyzers import thumbnails
+from conftest import TESTDATA, CollectionApiClient
+
+pytestmark = [pytest.mark.django_db]
+
+
+def test_thumbnail_job(fakedata, taskmanager, client):
+    test_doc = TESTDATA / './no-extension/file_doc'
+    with test_doc.open('rb') as f:
+        thumbnails.call_thumbnails_service(f, 100)
```
gabriel-v, line 11: Split this test into two:

- `test_thumbnail_service` that is this function
- `test_thumbanil_task` that does something else:
  - create a Blob object like here https://github.com/liquidinvestigations/hoover-snoop2/blob/229d8327b5508d7fe8e0ab82717dc6f2eaa7fbf2/testsuite/test_blobs.py#L9-L10
  - runs the Task function (the one annotated with @snoop_task) instead of the `call_thumbnails` service
  - checks that all your Thumbnail rows are there: `assert models.Thumbnail.objects.get(size=100, blob=X).thumbnail.size > 0`

None of these 2 tests requires `taskmanager` or `client` or `fakedata` since those are meant to be used directly


Here's another

snoop/data/analyzers/thumbnails.py
```diff
@@ -0,0 +1,375 @@
...
+        raise SnoopTaskBroken('thumbnail service returned http 500', 'thumbnail_http_500')
+
+    if (resp.status_code != 200
+            or resp.headers['Content-Type'] != 'image/jpeg'):
+        raise RuntimeError(f"Unexpected response from thumbnails-service: {resp}")
+    return resp.content
+
+
+@snoop_task('thumbnails.get_thumbnail')
+@returns_json_blob
```
gabriel-v, line 359: this doesn't return anything anymore, so remove the `@returns_json_blob` decorator (since it'll create a Blob with only `null` in it and set it as a result for all the function runs, wasting resources)

snoop/data/analyzers/thumbnails.py
```diff
@@ -0,0 +1,375 @@
...
+
+    Args:
+        data: the file for which a thumbnail will be created.
+        size: the size for the created thumbnail (thumbnail will be size x size)
+        """
+
+    url = settings.SNOOP_THUMBNAIL_URL + f'preview/{size}x{size}'
+    log.info(url)
+
+    session = requests.Session()
```
gabriel-v, line 344: Why do we need a new Session object here? Do you want to set any parameters on it?

If not, let's just do `requests.post(...)` to use the implicit session

snoop/data/analyzers/thumbnails.py
```diff
@@ -0,0 +1,375 @@
...
+
+    Returns the primary key of the created thumbnail blob.
+    """
+    SIZES = [100, 200, 400]
+
+    for size in SIZES:
+        with blob.open() as f:
+            resp = call_thumbnails_service(f, size)
+        blob_thumb = models.Blob.create_from_bytes(resp)
+        _, _ = models.Thumbnail.objects.get_or_create(
```
gabriel-v, line 371: This should be actually [update or create](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#update-or-create) because you want to always update the `thumbnail` while fetching the object based on `size` and `blob` (the stuff in your unique constraint.

So:
`models.Thumbnail.objects.update_or_create(blob=blob, size=size, defaults={'thumbnail': blob_thumb})`

snoop/data/analyzers/thumbnails.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+@snoop_task('thumbnails.get_thumbnail')
+@returns_json_blob
+def get_thumbnail(blob):
+    """Function that calls the thumbnail service for a given blob.
+
+    Returns the primary key of the created thumbnail blob.
+    """
+    SIZES = [100, 200, 400]
```
gabriel-v, line 365: Use the choices from your `models.Thumbnail.Size.choices` class 

Don't repeat yourself, what if tomorrow we want to add or change the sizes? We'd have to edit like 3-4 places in the code, not nice.

snoop/data/models.py
```diff
@@ -882,3 +882,32 @@ def __str__(self):
...
+
+class Thumbnail(models.Model):
+    """Database model for storing the Thumbnail corresponding to a Digest.
+    """
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=['blob', 'size'], name='unique_size')
+        ]
+
+    class Size(models.IntegerChoices):
```
gabriel-v, line 895: Call this class `SizeChoices` so it's more descriptive

snoop/data/models.py
```diff
@@ -882,3 +882,32 @@ def __str__(self):
...
+        SMALL = 100
+        MEDIUM = 200
+        LARGE = 400
+
+    blob = models.ForeignKey(
+        Blob,
+        on_delete=models.CASCADE,
+        related_name='file_blob',
```
gabriel-v, line 903: Actually, the related name is the name you'd use on the referenced table to get back to this one. So the related names here are incorrect, since the objects you get are going to be Thumbnail, not Blob.


 Since we don't want to get from any `Blob` back to the thumbnail by using some related name directly, let's deactivate it by setting it to `+`.

https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.ForeignKey.related_name

snoop/data/models.py
```diff
@@ -882,3 +882,32 @@ def __str__(self):
...
+    )
+
+    thumbnail = models.ForeignKey(
+        Blob,
+        on_delete=models.CASCADE,
+        related_name='thumbnail_blob',
```
gabriel-v, line 909: same as above: related_name +

PR #443, opened 2021-05-05
Pdf preview task

k-jell:
closes CRJI/EIC#678


testsuite/test_pdf_preview.py
```diff
@@ -0,0 +1,54 @@
+import pytest
+from snoop.data.analyzers import pdf_preview
+from conftest import TESTDATA, CollectionApiClient
+from snoop.data import models
+
+pytestmark = [pytest.mark.django_db]
+
+
+def test_pdf_preview_service():
+    test_doc = TESTDATA / './disk-files/word/sample-doc-file-for-testing-1.doc'
```
gabriel-v, line 10: put this filename in a const in this test module, since you're using it in a few tests

gabriel-v (commented):
lgtm, please rebase and fix conflicts. you may want to collapse your branch into a single commit first to only fix the conflicts once.

snoop/data/analyzers/pdf_preview.py
```diff
@@ -0,0 +1,96 @@
...
+            or resp.headers['Content-Type'] != 'application/pdf'):
+        print(resp.content)
+        raise RuntimeError(f'Unexpected response from pdf generator: {resp}')
+
+    return resp.content
+
+
+@snoop_task('pdf_preview.get_pdf')
+# the @returns_json_blob decorator is only needed to check if this function ran in digests.gather
+@returns_json_blob
```
gabriel-v, line 76: I don't think we need this; it creates overhead by writing out the result json `null` over and over again into some blob. If you only want to differentiate between running and TaskBrokenError, then I suspect you can just remove this decorator and expect to receive None as the value when it worked.

snoop/data/analyzers/pdf_preview.py
```diff
@@ -0,0 +1,96 @@
...
+"""
+
+
+def can_create(blob):
+    """Checks if the pdf generator can process this file."""
+    if blob.mime_type in PDF_PREVIEW_MIME_TYPES:
+        return True
+
+
+def call_pdf_generator(data, filename):
```
gabriel-v, line 56: Do we really need filename/extension for the service to work properly? Won't it work if we send it a filename of `a` and the correct mime type?

I don't want the complication of working with filenames here, because:
- documents can appear multiple times with different extensions, what is the correct one?
- the filenames themselves can be broken bytes that come from some outdated filesystem, so they're problematic to send around -- if we really need the extension, then we need to send the filename as `'a.' + extension` instead of the original filename.

snoop/data/analyzers/pdf_preview.py
```diff
@@ -0,0 +1,96 @@
...
+# the @returns_json_blob decorator is only needed to check if this function ran in digests.gather
+@returns_json_blob
+def get_pdf(blob):
+    """Calls the pdf generator for a given blob.
+
+    Adds the pdf preview to the database
+    """
+    filename = models.File.objects.filter(original=blob.pk)[0].name
+    filename_root, ext = os.path.splitext(filename)
+    if ext not in PDF_PREVIEW_EXTENSIONS:
```
gabriel-v, line 84: Same as above: do we really care about extensions at all? Couldn't we replace this with a mime type check? Then, we wouldn't even have an error condition, since we wouldn't be starting the task if the mime types don't match

gabriel-v (changes requested):
Small question on whether we really need to send filename to PDF preview service

snoop/data/analyzers/pdf_preview.py
k-jell: We do need to send a filename with extension, because that's how the service determines which conversion to run. But I replaced the actual filename with a default one. And the check before starting the task and processing the blob is using mime types already, based on the supported extensions.

snoop/data/urls.py
```diff
@@ -23,5 +23,6 @@
     path('<collection>/<hash>/raw/<filename>', views.document_download),
 
     path('<collection>/<hash>/tags/<username>/<uuid>', include(tags_router.urls)),
-    path('<collection>/<hash>/thumbnail/<size>.jpg', views.thumbnail)
+    path('<collection>/<hash>/thumbnail/<size>.jpg', views.thumbnail),
+    path('<collection>/<hash>/pdf-preview/', views.pdf_preview),
```
gabriel-v, line 27: no trailing slash please, check the patterns above this list: `.../json`

testsuite/conftest.py
```diff
@@ -183,3 +183,10 @@ def get_thumbnail(self, blob_hash, size):
         with mask_out_current_collection():
             resp = self.client.get(url)
         assert resp.status_code == 200
+
+    def get_pdf_preview(self, blob_hash):
+        col = collections.current()
+        url = f'/collections/{col.name}/{blob_hash}/pdf-preview/'
```
gabriel-v, line 189: no trailing slash, see above

gabriel-v (changes requested):

gabriel-v (approved):

PR #444, opened 2021-06-17
Image classification task

k-jell:
closes CRJI/EIC#730


snoop/data/analyzers/image_classification.py
```diff
@@ -0,0 +1,147 @@
...
+from django.conf import settings
+from ..tasks import SnoopTaskBroken, snoop_task, returns_json_blob
+import json
+import io
+from PIL import Image, UnidentifiedImageError
+
+
+PROBABILITY_LIMIT = 20
+
+"""Based on https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#image-file-formats"""
```
gabriel-v, line 14: docstrings for assignments should go after it, check the docs output at https://hoover-snoop2--444.org.readthedocs.build/en/444/reference/snoop/data/analyzers/image_classification/ -- the link goes under `PROBABILITY_LIMIT` which is not what you want

snoop/data/analyzers/image_classification.py
```diff
@@ -0,0 +1,147 @@
+"""Task to call a service that runs object detection and/or image classification on images"""
+
+import requests
```
gabriel-v, line 3: generally you want to sort imports like this:
- python std libs first, like json and io
- external libs second, like requiests, PIL, django
- project relative imports last, like ..tasks,  ..models

it doesn't usually matter, but it helps to be organized

testsuite/test_image_classification.py
```diff
@@ -0,0 +1,52 @@
...
+from conftest import TESTDATA, CollectionApiClient
+
+pytestmark = [pytest.mark.django_db]
+
+TEST_IMAGE = TESTDATA / './disk-files/images/bikes.jpg'
+
+
+def test_classification_service():
+    with TEST_IMAGE.open('rb') as f:
+        image_classification.call_image_classification_service(f, 'bikes.jpg')
```
gabriel-v, line 12: please check the results in this test, or at least check for some important classes

testsuite/test_image_classification.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def test_detection_service():
+    with TEST_IMAGE.open('rb') as f:
+        image_classification.call_object_detection_service(f, 'bikes.jpg')
```
gabriel-v, line 17: check results in this test too, and also in the following 2 tests

gabriel-v (changes requested):
LGTM, needs small changes around tests and docstrings

gabriel-v (approved):

gabriel-v (approved):

PR #471, opened 2022-07-18
Import tags

k-jell:
closes CRJI/EIC#876

### DIFF
--- a/snoop/data/management/commands/importtags.py
+++ b/snoop/data/management/commands/importtags.py
@@ -32,30 +32,33 @@ def read_csv(file_path):
     return taglist
 
 
-def update_tags(md5, tags, collection, uuid):
+def update_tags(md5, tags, collection, uuid, username, public=False):
     """Create document tags for all new tags from the list."""
+    print(public)
     collection = collections.ALL[collection]
     with collection.set_current():
         blob = models.Blob.objects.get(md5=md5)
         digest = models.Digest.objects.get(blob=blob.pk)
         existing_tags = models.DocumentUserTag.objects.filter(digest=digest.pk).values_list('tag', flat=True)
-        models.DocumentUserTag.objects.create(digest=digest, uuid=uuid, tag=' abc ',
-                                              user='kjell', public=False)
         for tag in tags:
             if tag not in existing_tags:
                 models.DocumentUserTag.objects.create(digest=digest, uuid=uuid, tag=tag,
-                                                      user='kjell', public=False)
+                                                      user=username, public=public)
 
 
 class Command(BaseCommand):
     "Import Tags UUIDs for all collections. JSON content is read from stdin."
 
     def add_arguments(self, parser):
-        parser.add_argument('collection', help='collection name')
-        parser.add_argument('taglist', help='Path to csv with tags.')
-        parser.add_argument('uuid', help='UUID of user for which the tags are imported.')
+        parser.add_argument('-c', '--col', help='collection name')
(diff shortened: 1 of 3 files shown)


snoop/data/management/commands/importtags.py
```diff
@@ -0,0 +1,72 @@
...
+    collection = collections.ALL[collection]
+    updated = False
+    with collection.set_current():
+        blob = models.Blob.objects.get(md5=md5)
+        digest = models.Digest.objects.get(blob=blob.pk)
+        existing_tags = models.DocumentUserTag.objects.filter(digest=digest.pk).values_list('tag', flat=True)
+        for tag in tags:
+            if tag not in existing_tags:
+                updated = True
+                models.DocumentUserTag.objects.create(digest=digest, uuid=uuid, tag=tag,
```
gabriel-v, line 47: use `get_or_create`, it returns a `object, created?` tuple so you can return that value.

Note: as it is, code is wrong, because the `existing_tags` query is not filtered by uuid and username.  This means you won't add some tags for some user, if another user has set the same tags.

snoop/data/management/commands/importtags.py
```diff
@@ -0,0 +1,72 @@
...
+                                                      user=username, public=public)
+                log.info(f'Created new tag: "{tag}" for document: "{md5}"')
+        return updated
+
+
+class Command(BaseCommand):
+    "Import Tags UUIDs for all collections. JSON content is read from stdin."
+
+    def add_arguments(self, parser):
+        parser.add_argument('-c', '--col', help='collection name')
```
gabriel-v, line 57: make long option `--collection`, i don't know about using this contraction anywhere else in the CLI

snoop/data/management/commands/importtags.py
```diff
@@ -0,0 +1,72 @@
...
+        parser.add_argument('-t', '--tags', help='Path to csv with tags.')
+        parser.add_argument('--uuid', help='UUID of user for which the tags are imported.')
+        parser.add_argument('--user', help='Username matching the UUID.')
+        parser.add_argument('-p', action='store_true', help='Flag to set the tags as public.')
```
gabriel-v, line 61: add long option `--public`

testsuite/test_tagimport.py
```diff
@@ -0,0 +1,89 @@
...
+    digest1 = models.Digest.objects.get(blob_id=blob1.pk)
+    digest2 = models.Digest.objects.get(blob_id=blob2.pk)
+
+    assert models.DocumentUserTag.objects.filter(digest_id=digest1.pk, tag='tag1').exists()
+    assert models.DocumentUserTag.objects.filter(digest_id=digest1.pk, tag='tag2').exists()
+    assert models.DocumentUserTag.objects.filter(digest_id=digest2.pk, tag='tag1').exists()
+    assert models.DocumentUserTag.objects.filter(digest_id=digest2.pk, tag='tag3').exists()
+
+    # need to wait for the document to be indexed
+    time.sleep(5)
```
gabriel-v, line 49: this will make for flaky tests, that depend on machine load and other things we can't control.

Do a polling loop on the API for max. 2min, it will tell you when the tag has been indexed.

gabriel-v (changes requested):

gabriel-v:
Also: do we really need to specify the UUID in the CLI?  I think we can get it from Search using the user profile api call, right? That would also take care of username validation: if a missing username is specified, we can't get the UUID either, and it will fail.

gabriel-v (approved):

PR #485, opened 2023-06-08
Webdav mounting

k-jell:
closes CRJI/EIC#1014


k-jell:
old branch, for recent version see #489

PR #488, opened 2023-09-14
Dynamic collection

k-jell:


snoop/data/collections.py
```diff
@@ -406,6 +432,8 @@ def allow_migrate(self, db, app_label, model_name=None, **hints):
         Snoop models not allowed in 'default'; other models not allowed
         in collection_* databases
         """
+        if model_name == 'nextcloudcollection':
```
gabriel-v, line 435: remove this, the model is in another app now

snoop/data/collections.py
```diff
@@ -387,10 +406,17 @@ class CollectionsRouter:
...
+    # TODO change router so that it doesn't have to know which model goes where
+    # make a new base class for collection routed models
+
     snoop_app_labels = ['data']
 
     def db_for_read(self, model, instance=None, **hints):
+        """NextcloudCollection Table needs to go into default db."""
+
         if model._meta.app_label in self.snoop_app_labels:
+            if model._meta.object_name == 'NextcloudCollection':
```
gabriel-v, line 418: remove this, the model is in another app now

snoop/data/collections.py
```diff
@@ -387,10 +406,17 @@ class CollectionsRouter:
     Uses the current collection's `.db_alias` to decide what database to route the reads and writes to.
     """
 
+    # TODO change router so that it doesn't have to know which model goes where
+    # make a new base class for collection routed models
+
     snoop_app_labels = ['data']
 
     def db_for_read(self, model, instance=None, **hints):
+        """NextcloudCollection Table needs to go into default db."""
```
gabriel-v, line 415: remove this, the model is in another app now

gabriel-v (changes requested):

snoop/data/admin.py
```diff
@@ -1296,7 +1296,7 @@ def get_admin_links():
 
 DEFAULT_ADMIN_NAME = '_default'
 sites = {}
-for collection in collections.ALL.values():
+for collection in collections.get_all(static=True):
```
gabriel-v, line 1299: this `static=` arg should be called `static_only` because setting it to False still gives you static entries

snoop/defaultsettings.py
```diff
@@ -546,3 +546,10 @@ def bool_env(value):
     SENTRY_DSN = os.getenv('SENTRY_DSN')
     SENTRY_SAMPLE_RATE = float(os.getenv('SENTRY_SAMPLE_RATE', '1.0'))
     sentry_sdk.init(dsn=SENTRY_DSN, traces_sample_rate=SENTRY_SAMPLE_RATE, shutdown_timeout=0)
+
+ENABLE_DYNAMIC_COLLECTION_ADMINS = os.getenv('ENABLE_DYNAMIC_COLLECTION_ADMINS', False)
```
gabriel-v, line 550: do `bool(os.getenv(..., ''))` instead, so this always evaluates to bool

snoop/data/collections.py
```diff
@@ -323,20 +341,21 @@ def drop_db(db_name):
 
 
 def create_databases():
-    """Go through [snoop.data.collections.ALL][] and create databases that don't exist yet."""
+    """Go through [snoop.data.collections.list_keys()][] and create databases that don't exist yet."""
 
     dbs = all_collection_dbs()
-    for col in ALL.values():
+    for col_name in list_keys():
```
gabriel-v, line 347: simplify into ` for col in get_all():` and use `col.name` instead of `col_name`?

gabriel-v (changes requested):

snoop/data/collections.py
```diff
@@ -73,35 +71,55 @@ class Collection:
...
         variant_list = [[col.strip() for col in variant.split(':')] for variant in table_headers.split(';')]
-        self.default_table_head_by_len = {
+        return {
             len(variant): variant
             for variant in variant_list
             if len(variant) > 1
         }
+
+    @property
+    def explode_table_rows(self):
```
gabriel-v, line 122: accident happened here : you probably want `return` instead of that

snoop/common_data/models.py
```diff
@@ -36,3 +41,52 @@ class Meta:
...
+                            f'{settings.SNOOP_WEBDAV_MOUNT_DIR}'
+                            f'/{self.name}/data'
+                        )
+                        ], check=True)
+        secrets_content = (
+            f'{settings.SNOOP_WEBDAV_MOUNT_DIR}'
+            f'/{self.name}/data'
+            f' {self.webdav_user} {self.webdav_password}'
+        )
+        with open('/etc/davfs2/secrets', 'a') as secrets_file:
```
gabriel-v, line 79: 1. if you open this in append mode, how do you know you won't get duplicate entries every time you run this function?
2. if this gets executed in parallel by many workers, how do you know this file won't be corrupted?
3. can we move this into a separate file for each collection, that you put a file lock around?
4. is there a method for removing old/unused mounts? We don't want to have a million active mounts around all the time

snoop/common_data/models.py
```diff
@@ -36,3 +41,52 @@ class Meta:
...
+            secrets_file.write(f'\n{secrets_content}')
+            mount_command = (
+                f'mount -t davfs http://10.66.60.1:9972{self.webdav_url} '
```
gabriel-v, line 82: please move http://10.66.60.1:9972 into setting and environment variable (default '') and assert that it's set when running the mount operation.

then make another PR in node, to set the correct env var (passed through a fabio port and not a static port) so it still works in multi-node environments

gabriel-v (changes requested):

PR #489, opened 2023-09-28
Nextcloud collections

k-jell:
closes https://github.com/CRJI/EIC/issues/1014

### DIFF
--- a/snoop/data/s3.py
+++ b/snoop/data/s3.py
@@ -255,7 +255,6 @@ def get_webdav_mount(mount_name, webdav_username, webdav_password, webdav_url):
     Returns:
         The path in the filesystem where the webdav share is mounted.
     """
-
     paths = get_paths(mount_name)
     target_path = os.path.join(paths.get('target_path'), 'data')
 
@@ -304,7 +303,7 @@ def write_password_file(path, content):
     """
     with open(path, 'wb') as pass_temp:
         password_file = pass_temp.name
-        subprocess.check_call(f"chmod 600 {password_file}", shell=True)
+        subprocess.check_call(['chmod', '600', password_file])
         pass_str = content
         pass_temp.write(pass_str)
         pass_temp.close()
@@ -448,8 +447,8 @@ def adjust_mounts(mount_name, old_info,
     # create new mount
     logger.info('creating new mount: %s', mount_name)
     clean_makedirs(mount_args.get('target_path'))
-    pid = mount_func(mount_args)
 
+    pid = mount_func(mount_args)
     # create new entry with pid and timestamp
     new_info[mount_name] = {
         'pid': pid, 'timestamp': timestamp(),
--- a/snoop/data/s3.py
+++ b/snoop/data/s3.py
@@ -172,7 +172,7 @@ def clear_mounts():
                 except Exception as e:
                     logger.debug('clear mounts info corrupted: %s', e)
         for mount_value in mount_info.values():
-            umount_s3fs(mount_value['target'], mount_value['pid'])
+            umount(mount_value['target'], mount_value['pid'])
         os.remove(mount_info_path)
     # in case json is kaputt, let's use `find` to try to unmount all targets
     targets = subprocess.check_output(
@@ -186,7 +186,7 @@ def clear_mounts():
         target = target.strip()
         if target:
             try:
-                umount_s3fs(target)
+                umount(target)
             except Exception as e:
                 logger.warning('could not umount s3fs! %s', str(e))
                 raise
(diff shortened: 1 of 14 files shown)


snoop/data/s3.py
```diff
@@ -421,10 +550,27 @@ def _data_mounted():
...
-    raise RuntimeError(f'cannot remove old S3 mounts! target={target}, pid={pid}')
+    tracer.count("umount_failed")
+    raise RuntimeError(f'cannot remove old mounts! target={target}, pid={pid}')
+
+
+def mount_webdav(mount_args):
+    """Run subprocess to mount webdav share to target. Will wait until completed."""
+
+    cmd_bash = f"""
+    mount -t davfs http://10.66.60.1:9972{mount_args.get("webdav_url")} \\
```
gabriel-v, line 566: Left-over hardcoded IP/PORT, replace with new env

Also run `rg 10.66.60.1`  to find any other hard-coded ips we missed

snoop/data/tasks.py
```diff
@@ -1788,3 +1789,29 @@ def lock_table(model):
...
+    search_url = settings.HOOVER_SEARCH_URL + 'nextcloud_collections'
+    json = requests.get(search_url).json()
+    for nc_col in json:
+        col_name = nc_col.get('name')
+        if not common_models.NextcloudCollection.objects.filter(name=col_name).exists():
+            col = common_models.NextcloudCollection(name=nc_col.get('name'),
+                                                    opt=nc_col
+                                                    )
+            col.save()
+            print(f'Created collection {col_name}.')
```
gabriel-v, line 1808: replace all prints with logs

snoop/defaultsettings.py
```diff
@@ -553,3 +563,12 @@ def bool_env(value):
 to hit the database, which is not available during that stage
 of the CI build.
 """
+
+SNOOP_WEBDAV_MOUNT_DIR = os.path.join(os.getenv('MNT', '/mnt'), 'snoop-webdav-mounts', 'collections')
```
gabriel-v, line 567: I think these `SNOOP_WEBDAV_CONF_DIR` `SNOOP_WEBDAV_MOUNT_DIR`  `SNOOP_WEBDAV_SECRETS_DIR` are left-overs / dead code. Let's remove them, since our webdavs are going to be mounted in the same place as s3 (in the custom temp dir), we don't want any reference to `/mnt` or defaults into `/opt/hoover/snoop`

snoop/data/s3.py
```diff
@@ -278,9 +408,8 @@ def target_is_mounted(path):
     )
 
 
-def adjust_s3_mounts(mount_name, old_info,
-                     bucket, mount_mode,
-                     password_file_path, address, target_path, logfile_path):
+def adjust_mounts(mount_name, old_info,
+                  mount_func, mount_args):
```
gabriel-v, line 412: Add new boolean argument `check_pid` that is True on S3 and False on webdav.

Then, at line 434, replace
```
new_info = _clear_dead(old_info)
```

with

```
if check_pid:
    new_info = _clear_dead(old_info)
```


This should fix the problem with non-foreground mounting.

snoop/data/s3.py
```diff
@@ -421,10 +550,27 @@ def _data_mounted():
...
+
+
+def mount_webdav(mount_args):
+    """Run subprocess to mount webdav share to target. Will wait until completed."""
+
+    cmd_bash = f"""
+    mount -t davfs http://10.66.60.1:9972{mount_args.get("webdav_url")} \\
+        -o conf={mount_args.get("configfile_path")} \\
+        {mount_args.get("target_path")} > {mount_args.get("logfile_path")} \\
+        2>&1 & echo $!
```
gabriel-v, line 569: Remove the `& echo $!` at the end, because this mount command will not be running in foreground, so we want to wait for it to finish.

You can set the pid to -1 (since we won't check for it's validity, see other comment in `adjust_mounts`

snoop/data/views.py
```diff
@@ -724,3 +725,10 @@ def processing_status(request, hash):
     if total_count != 0 and result['done_count'] == total_count:
         result['finished'] = True
     return JsonResponse(result)
+
+
+def get_nextcloud_collections(request):
```
gabriel-v, line 730: let's put this view in the common app, since it only accesses the common models.

gabriel-v (changes requested):
Looks very good - very little left to fix.

Also, please look at the mkdocs build, it's failing (probably newly installed webdav libraries need to be ignored in the docs ignore file?)

snoop/defaultsettings.py
```diff
@@ -553,3 +563,6 @@ def bool_env(value):
 to hit the database, which is not available during that stage
 of the CI build.
 """
+
+SNOOP_NEXTCLOUD_URL = os.getenv('SNOOP_NEXTCLOUD_URL')
+"""Location on disk where webdav mounts are stored."""
```
gabriel-v, line 568: the url is not a "Location on disk where"

snoop/data/s3.py
```diff
@@ -421,10 +554,26 @@ def _data_mounted():
...
+    """Run subprocess to mount webdav share to target. Will wait until completed."""
+
+    cmd_bash = f"""
+    mount -t davfs {settings.SNOOP_NEXTCLOUD_URL}{mount_args.get("webdav_url")} \\
+        -o conf={mount_args.get("configfile_path")} \\
+        {mount_args.get("target_path")} > {mount_args.get("logfile_path")} \\
+        2>&1
+    """
+    logger.info('running davfs process: %s', cmd_bash)
+    tracer.count("mount_webdav_start")
```
gabriel-v, line 576: where did the `subprocess.check_call` go? This doesn't mount anything now

gabriel-v:
plz fix associated error in master:

https://jenkins.liquiddemo.org/liquidinvestigations/node/4029/2/4

```
err:   File "/opt/hoover/snoop/snoop/data/management/commands/resetcollectionindex.py", line 19, in handle
err:     with Collection(collection).set_current():
err: TypeError: Collection() takes no arguments
```

